PR #3, opened 2025-03-10
React source code parsing using treesitter and Rust

turizspace:
working test in ast/src/testing/react
Screenshot 2025-03-10 023056


Evanfeenstra:
Very nice! But for me the test doesn't pass.

I get 41 edges, fixed `NewPerson` file path, and 5 Calls. weird i wonder why...

turizspace:
It’s due to assertion mismatch of excepted values. fixing  my ennviroment

turizspace:
Screenshot 2025-03-10 075319

Evanfeenstra:
why does NewPerson have a backslash and only on one place?

turizspace:
path is getting generated with backslashes (\\), and others are normalized with forward slashes (/) cus im using Windows

Evanfeenstra:
ok normalize them all somehow

turizspace:
Screenshot 2025-03-10 080926
the error when i use path as is , the back slash solves it

turizspace:
ok let me normalize them

turizspace:
added function to normalize paths and replace backslashes with forward slashes

Evanfeenstra:
all your files have git symbols on top, like in the text itself

please remove the endpoint_finders in react_ts.rs, frontend doesnt need to define endpoints

turizspace:
updated.

Evanfeenstra:
can u make a .gitignore with node_modules and package-lock.json

Evanfeenstra:
i get 44 nodes and 44 edges, so the test still doesn't pass for me

Why are we getting different results? Are you able to run the full test? Maybe the backslash is changing something?

turizspace:
https://github.com/user-attachments/assets/9bd20728-be45-45e1-a9dd-53d47541be96

it is running successfully.  similar to before when you had got 41 edges.

PR #4, opened 2025-03-10
feature: created an agnostic server code parser

fayekelmith:
## Task
Create an agnostic test util, `test_backend,` that can run server code in different programming languages. 
In this PR the test_backed tests the following in a server:
- [X] Language node
- [X] package file
- [X} 2 endpoints: GET "person/:id" and POST "person"
- [X} 1 Datamodel called "Person" with each of these two functions having a Contains edge to the Person Datamodel OR the Call another Function which Contains the Datamodel

## Loom Video
https://www.loom.com/share/e9f767bf06fb42e38d0b924bcd42542e?sid=0b19d64f-dfdb-436f-8ce2-3dbce8e5c51c

## Bounty: 
https://community.sphinx.chat/bounty/3970


fayekelmith:
@Evanfeenstra, please, sir, should I replace the `test_backend` utill in the go and python mod.rs files?

Secondly, I would like to request you watch the Loom video and understand where I think the `ast/src/lang/queries/go.rs Line 243` file could use some improvement with the `data_model_within_query`. 

I believe that the query is not comprehensive enough to obtain all the ways in which the Data Model was used in the Endpoint, hence why we fail to get a pass in the last test case with the Go tests. 

So, my requests are:
1. Should I implement `test_backend` directly in the python and go code?  (after you review and I improve)
2. Should I enhance the `ast/src/lang/queries/go.rs L243`  with more a more comprehensive query as i did with `ast/src/lang/queries/python.rs L131`  ?

Evanfeenstra:
1. No keep it in the `test_backend` file. That file is not very maintainable though... we will add many more tests later. So for example `test_endpoints` should take a `Vec<(String, String)>` argument (like a list of (endpoint,verb) pairs. And it checks each automatically. All the tests in that file can be refactored. 

In fact its probably best to create a struct:
```
struct BackendTester {
    graph: Graph,
    lang: Lang,
}
```

And implement the different tests as methods. So you dont need to pass "graph" and "lang" everywhere. Please create util methods like `find_endpoint`, so you don't need to repeat code so much!

2. No.. the Go one more tricky since the `Person` struct is not in the handler directly. Instead there is a Calls edge to the other DB function. I see now it is too confusing because they have the same name... I just pushed and updated their names. 

So there needs to be a recursive loop in `test_handler_functions` to check for Datamodel, and if it doesn't exist, check for Calls to another Function and then check that one too...

fayekelmith:
Alright sir... I'll get to improving the code quality right away.

fayekelmith:
@Evanfeenstra, I have refactored the code as you suggested and improved the recursive implementation that you suggested, sir. but it seems to fail. Perhaps you could suggest what I'm missing out on? I'll really appreciate it, sir.

ast/src/testing/test_backend.rs L238

Evanfeenstra:
Please do not edit the ast/src/testing/go/mod.rs file, you can always add a new test. 

The test_backend file still needs a lot of work. `test_backend` should be at the top. And each test should take arugments:

```
pub fn test_backend(&self) -> Result<()> {
        self.test_language_node()?;
        self.test_package_file()?;
        
        let data_model = "Person";
        self.test_data_model(data_model)?;

       let expected_endpoints = vec![("GET", "person/:param"), ("POST", "person")];
        self.test_endpoints(expected_endpoints)?;
        
        self.test_handler_functions(expected_endpoints, data_model)?;

        Ok(())
  }
  ```
  
Like that. So we can easily add new endpoints or whatever later on. There will be many more things to test, thats why all the stuff needs to be well defined in one place.

I think for now, just change the Python code to have a Database class that has GetPersonById and NewPerson methods. That way you can do the same test for Go/Python. The Nodes/Edges are: `Endpoint - HANDLER -> Function - CALLS -> Function - CONTAINS -> Datamodel`

PR #7, opened 2025-03-12
chore: filter out struct from class in Go

tobi-bams:


Evanfeenstra:
Cool! To keep things clean, and avoid lots of if statements in builder.rs, maybe we can implement it as a trait method instead. Like a `clean_graph` method in ast/src/lang/queries/mod.rs, that takes a `&mut Graph` as an argument, but does nothing. Than in the queries/go.rs file you can implement the filter

tobi-bams:
Hi @Evanfeenstra I have pushed the changes!

Evanfeenstra:
perfect!

PR #9, opened 2025-03-13
chore: fixed package file parsing for python and ts function parsing

fayekelmith:
This PR implements the following changes:
- fixes python package file errors and the inclusion of unnecessary files (.txt files) 
- sorts typescript function nodes so the intented functions are always picked


fayekelmith:
Requirements wasn't identified last time because python was commented out as one of the languages in the array for programming languages... 

If I uncomment it, it will permit python's package file to be identifiable which solves the problem without picking up every .txt file. 

This works similar to the other languages e.g in the Go files we didn't mention the extension for its package file but it was picked (because it's part of the programming languages in the array)

PR #10, opened 2025-03-14
rust backends and rust treesitter query

fayekelmith:
##  Objective
- [X] Set up Rust Backend with Axum
- [x] Configure rust Treesitter parsing

### Screenshot
Screenshot from 2025-03-20 02-50-21

### task
https://community.sphinx.chat/p/cuvb997paij6rm5nbasg/assigned/4038/0


fayekelmith:
@Evanfeenstra Greetings sir, Please, sir, I have set up the rust Axum server which works successfully  but I have a hard time putting together an `endpoint_finder` Treesitter query that works. Please do you have any recommendations ?

fayekelmith:
> Here's something that works. It might make false positives though, we may need to make it more specific if we find it matches functions that aren't actually endpoints
> 
> ```rust
> (call_expression
>     (arguments
>         (string_literal) @endpoint
>         (call_expression
>             function: (identifier) @verb (#match? @verb "^get$|^post$|^put$|^delete$")
>             arguments: (arguments
>                 (identifier) @handler
>             )
>         )
>     )
> ) @route
> ```

Worked Like Magic 🪄 
Thank you sir. I'll try to improve it to be more effective.

fayekelmith:
@Evanfeenstra, I wanted to ask you a few questions about this PR. I'm afraid I'm a little stuck again :( at the level of testing for the presence of the data model in the endpoint handler. I tried reaching out via the Sphinx app, but it seems to be down for the moment.

Evanfeenstra:
The data model is the in the db function such as `get_person_by_id`. For that the query can be very simple: any type in that function:
```
(type_identifier) @struct_name
```

`function_call_query` is not so easy... here is one that will work for many kinds:
```
(call_expression
    function: [
        (identifier) @FUNCTION_NAME
        ;; module method
        (scoped_identifier
            path: (identifier) @PARENT_NAME
            name: (identifier) @FUNCTION_NAME
        )
        ;; chained call
        (field_expression
            field: (field_identifier) @FUNCTION_NAME
        )
    ]
    arguments: (arguments) @ARGUMENTS
) @FUNCTION_CALL
```

fayekelmith:
@Evanfeenstra, please review this pr

Evanfeenstra:
Your queries are too complex. You can look in the parse.rs file to see how Nodes are formatted. For example, Datamodel is formatted with the `format_data_model` function. It only includes two constants: STRUCT_NAME and STRUCT. So the `data_model_query` should only include those 2 constants:
```
(struct_item
    name: (type_identifier) @STRUCT_NAME
) @STRUCT
```

Evanfeenstra:
imports query:
```
(source_file
    (use_declaration)+ @IMPORTS
)
```

Evanfeenstra:
`trait_query`:
```
(trait_item
    name: (type_identifier) @TRAIT_NAME
    body: (declaration_list)
) @TRAIT
```

fayekelmith:
@Evanfeenstra, I've updated the queries, and the tests pass now.
I appreciate your help.

PR #12, opened 2025-03-17
feature: setup node server with typeorm, sequelize and prisma

fayekelmith:
- [X] Express server 
- [X] Treesitter queries


ast/src/lang/queries/typescript.rs
```diff
@@ -112,97 +112,24 @@ impl Stack for TypeScript {
...
+                    object: (identifier) @app_object
+                    property: (property_identifier) @endpoint-verb (#match? @endpoint-verb "^get$|^post$|^put$|^delete$")
+                )
+                arguments: (arguments
+                    (string (string_fragment) @endpoint)
+                    [
+                    (arrow_function) @handler
+                    (function_expression) @handler
+                    (identifier) @handler
+                    ]
```
fayekelmith, line 128: please sir, @Evanfeenstra How can I improve these queries? Or are they even the problem? 
I'm not able to get the endpoints to be tracked, every other node is tracked accurately.

### Loom
https://www.loom.com/share/9df60b28a80f41a0b4ac4be1c408f8f7?sid=9e82d53e-bc17-461d-a646-0339208071d7

Evanfeenstra:
In the other languages we have done so far, the endpoint definition points to the identifier of the handler function. Here's how to do it in TS:
```ts
import express from "express";
import { PersonController } from "./controller.js";

export function registerRoutes(app) {
  app.get("/person/:orm/person/:id", getPerson);

  app.post("/person/:orm/person", postPerson);
}

function getPerson(req, res, next) {
    PersonController.getById(req, res).catch(next);
}

function postPerson(req, res, next) {
    PersonController.create(req, res).catch(next);
}
```

But then you need this to be covered in your function_definition_query too:
```
(function_declaration
  name: (identifier) @function-name
) @function-definition
```

fayekelmith:
@Evanfeenstra please review the updates.

Evanfeenstra:
Why does `new_typescript_for` have that Option arg? its not used anywhere

Evanfeenstra:
Also please add `jsx` to the from_str impl as well
```rs
impl FromStr for Lang {
    type Err = anyhow::Error;
    fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s {
            "tsx" | "jsx" => Ok(Lang::new_react_ts()),
            _ => {
                let ss = Language::from_str(s)?;
                Ok(Lang::from_language(ss))
            }
        }
    }
}
```

fayekelmith:
Resovled both concerns and introduced `LANGUAGE::REACT`

PR #18, opened 2025-03-21
[fix] package file duplication

fayekelmith:
### Task
- [X] fix issue with package file duplicacy

Ref
https://community.sphinx.chat/p/cuvb997paij6rm5nbasg/assigned/4101/0


fayekelmith:
@Evanfeenstra, fixed issue by removing file duplication when building the graph. 
 All non-excluded files were added and then package files were especially added, which affects how the graph is built for all languages.

PR #19, opened 2025-03-22
setup test_frontend

fayekelmith:
### Task
Setup test_frontend similar to test_backend. 

Testing:
- [X] Language
- [X] Package file
- [X] Data Model or Schema
- [X] Component
- [X] Pages
- [X] API requests


fayekelmith:
@Evanfeenstra please check this PR

Evanfeenstra:
Cool! Its a little too specific in a couple place though. Try to make it more general so it works for other languages next (swift, kotlin, etc)

- `.skip(3)` why 3? What if the route is different?
- `let repo = Some(server.to_string() + "_ts");` this wont work for all of course. I guess we can just change all the "react_ts" paths to "react" maybe...

Try to simplify in any way possible! I love the `verify_nodes` function, but why does it need a `get_name` function? That will always be the same for every node right?

fayekelmith:
@Evanfeenstra I have updated the logic for checking the endpoints and removed `get_name` you were right, sir. i didn't realise it was same throughout. 

also replaced `react_ts` with `react` to make it more generic.

PR #21, opened 2025-03-22
adding prisma query to data model detection

fayekelmith:
Attempt to add prisma data model query to typescript


fayekelmith:
@Evanfeenstra, please what do you think about this approach of adding Prisma querying to the mix? 

I've not successfully queried the data model yet, but if the tree-sitter query for prisma (https://crates.io/crates/tree-sitter-prisma-io) works then, its possible. 

another reference:
https://github.com/victorhqc/tree-sitter-prisma/blob/main/queries/highlights.scm

Evanfeenstra:
Cool! So by using build_graphs, you are getting both languages!

Please do not change the `new_multi_detect` argument, that is used already for real repos (each language has a default use_lsp setting)

Also, data model query needs the STRUCT_NAME too right?

fayekelmith:
Yes sir @Evanfeenstra, the approach is to build graphs for different languages.


I've tried different variants of the query (including some with STRUCT_NAME), without success, sir. 

And I'll revert changes you highlighted for new_multi_detect and keep trying.

fayekelmith:
@Evanfeenstra, this Prisma query now works. 

However, I would like to change the implementation approach. 

I found these methods in `ast/src/queries/mod.rs` :
 ```rust
fn use_data_model_within_finder(&self) -> bool {
        false
    }
    fn data_model_within_finder(&self, _dm: &NodeData, _graph: &Graph) -> Vec<Edge> {
        Vec::new()
    }
```

but they are not implemented in `ast/src/lang/parse.rs` at line 479 - 482
```rust
 if self.lang.use_data_model_within_finder() {
                    // do this later actually
                    // models = self.lang.data_model_within_finder(&func.name, file, graph);
        }
```

I was wondering what the purpose of these methods was and if I can use them instead, in the TypeScript implementation to:
- locate the prisma file, `schema.prisma`
- parse it
- query the data models from it 
- process each of them and add them to the graph. 

Please do this sound like a plausible approach?  
and please those methods meant for something else or can they be used as such?

fayekelmith:
@Evanfeenstra, I used the approach described above with a different method (one I defined). 
 
 Please review the PR.

fayekelmith:
The Prisma Treessiter Libraries are now incompatible with our current Treesitter version.

PR #23, opened 2025-03-25
Svelte source code parser

turizspace:
Added svelte test and queries


Evanfeenstra:
frontend language like svelte needs `request_finder` (look at react for example, should be the same)

But it does NOT need the endpoint_finders, that is for backend only

PR #24, opened 2025-03-26
more front-end testing with `test_frontend`

fayekelmith:
- [x] Improve Swift Queries
- [x] Improve Kotlin Queries
- [x] Integrate with test_frontend

### DIFF
--- a/ast/src/lang/queries/swift.rs
+++ b/ast/src/lang/queries/swift.rs
@@ -138,34 +138,35 @@ impl Stack for Swift {
             }
         }
     }
+
     fn data_model_query(&self) -> Option<String> {
         Some(format!(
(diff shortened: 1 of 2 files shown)


fayekelmith:
@Evanfeenstra, sir, I had to hardcode the data model query for Swift. 
I tried getting it correctly using commented out query but to no avail. Please do you can you point out something I may be missing?

Resource used:
https://swift-ast-explorer.com/
https://github.com/alex-pinkus/tree-sitter-swift/blob/main/queries/highlights.scm

Screenshot from 2025-03-27 03-58-09

fayekelmith:
@Evanfeenstra, please review this PR

PR #31, opened 2025-04-01
refactored graph into traits and implementation

fayekelmith:
- [X] created trait out of graph methods
- [X] refactored helper methods for graph methods

### DIFF
--- a/ast/examples/file.rs
+++ b/ast/examples/file.rs
@@ -1,9 +1,9 @@
 use anyhow::Result;
+use ast::lang::ArrayGraph;
+use ast::utils::logger;
 use ast::Lang;
 use ast::{self, repo::Repo};
 use lsp::language;
-use ast::utils::logger;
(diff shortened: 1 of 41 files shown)


Evanfeenstra:
also `process_endpoint_groups` should be added to the Graph

Evanfeenstra (commented):
- please minimize changes. Just make `build_graph` use `ArrayGraph` directly, so the caller does not need to choose. So all those example files don't need to change
- Try to remove `nodes_mut` and `edges_mut`. We want to avoid returning `Vec`. So instead make `extend_nodes` and `extend_edges` methods. Instead of `graph.edges().push`, do `graph.push_edge()`
- Why are you removing functions like `find_funcs_by` or `find_class_by`? We don't want to iterate nodes in order to find one. It will be much faster to use a Map, that's why we are making it generic.

Issue #38, opened 2025-04-03
Graph trait for abstract nodes store


Evanfeenstra:
Graph trait with the method:

​`find_nodes(node_type: NodeType, name: &str) -> Vec<Node>​`

This we can use in many places. For example, instead of the `find_only_one_function_file(func_name: &str, graph: &Graph)` function in ast/src/lang/parse.rs, we can add another method to Graph, which uses `self.find_nodes` internally. 

Sometimes its good to avoid a generic arg, like for `find_function_parent`. So try a closure arg intead:

`find_function_parent(..., class_finder: CF) -> Result<Option<Operand>> where CF: Fn(&str) -> Option<NodeData>`.

PR #40, opened 2025-04-03
refactor to graph trait

fayekelmith:
- [ ] Create Graph Trait and Refactor Codebase to implement it. 

- [X] Step One
Referenceing : https://github.com/stakwork/stakgraph/pull/31/files#pullrequestreview-2738458784
And the the two critical methods `find_node` and `find_nodes` : #38 

@Evanfeenstra, is this a good base?


Evanfeenstra:
Simplify even more. For now, only one method `find_nodes`. It always takes a node_type (no Option)

PR #46, opened 2025-04-03
Abstacting Graph Traits

fayekelmith:
Abstracting Graph Traits

### DIFF
--- a/ast/src/lang/graph.rs
+++ b/ast/src/lang/graph.rs
@@ -93,34 +93,34 @@ impl NodeRef {
 }
 
 impl Graph for ArrayGraph {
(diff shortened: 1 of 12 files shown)


Evanfeenstra:
find_node_at, not find_data_model_at

Make it accept a node_type

Evanfeenstra:
do not pass Nodes as arg to find_function_parent

Pass a callback like this: `&dyn Fn(&str) -> Option<NodeData>`. Then in the caller (`format_function`) you can you the `graph`

Evanfeenstra:
Great! 

I actually think its cleaner to have the Graph method return NodeData instead of Node.

Then its easier to use it elsewhere

PR #48, opened 2025-04-04
converting the `parse.rs` file to be Graph Trait compatible

fayekelmith:
- [ X] format endpoints

### DIFF
--- a/ast/src/lang/graph.rs
+++ b/ast/src/lang/graph.rs
@@ -143,7 +143,7 @@ impl Graph for ArrayGraph {
             }
         })
     }
-    fn find_node_by_name_and_file_ends_with(
+    fn find_node_by_file_ends_with(
         &self,
         node_type: NodeType,
         name: &str,
@@ -160,7 +160,7 @@ impl Graph for ArrayGraph {
             }
         })
     }
-    fn find_nodes_by_name_and_file_ends_with(
+    fn find_nodes_by_file_ends_with(
         &self,
         node_type: NodeType,
         name: &str,
--- a/ast/src/lang/graph.rs
+++ b/ast/src/lang/graph.rs
@@ -160,6 +160,22 @@ impl Graph for ArrayGraph {
             }
         })
     }
+    fn find_nodes_by_name_and_file_ends_with(
+        &self,
+        node_type: NodeType,
+        name: &str,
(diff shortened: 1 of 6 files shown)


Evanfeenstra:
Only one change: review the original implementation of the general controller:
```
let controllers =
                graph.find_funcs_by(|nd: &NodeData| is_controller(nd, handler_string));
            debug!(
                "ror endpoint controllers for {}: {:?}",
                handler_string,
                controllers.len()
            );
            for nd in controllers {
                debug!("checking controller: {}", nd.name);
                if actions.contains(&nd.name) {
                    debug!("===> found action: {}", nd.name);
                    let mut endp_ = endpoint.clone();
                    endp_.add_action(&nd.name);
                    if let Some(verb) = verb_map.get(&nd.name) {
                        endp_.add_verb(verb);
                    }
                    inter.push((endp_, nd));
                }
            }
            ```
            
You have this:
```
for action in actions {
                if let Some(nd) = find_handler(&action, &CONTROLLER_FILE_SUFFIX) {
                    debug!("===> found action: {}", nd.name);
                    let mut endp_ = endpoint.clone();
                    endp_.add_action(&nd.name);
                    if let Some(verb) = verb_map.get(&nd.name) {
                        endp_.add_verb(verb);
                    }
                    inter.push((endp_, nd));
                }
            }
            ```
            
While the end result is the same, its less efficient. When we introduce a DB here instead of just an in-memory graph, it will cause 7 DB requests for every general controller in Ruby, instead of 1.

So please add a `find_node_by_file_ends_with` trait method, and use it in that one place

fayekelmith:
@Evanfeenstra, I have implemented the fixes.

Evanfeenstra:
Why did you add `find_nodes_by_name_and_file_ends_with`? You need `find_nodes_by_file_ends_with`

Look at the original code snippet above. It finds ALL functions whose `file` ends_with a certain string. There is no "name" in the query. And then afterward, it find matches by names...

Evanfeenstra:
Great! all the logic looks good! Can you just rename the callbacks...

`callback`/`special_callback` in lang/queries/mod.rs and in lang/queries/ruby.rs are too general. Can you call them `find_fn` and `find_fns_in`.

PR #50, opened 2025-04-04
implementing Graph Trait in build.rs

fayekelmith:
- [X] refactor methods of the form `add_<element>`
- [x] handle special cases

### DIFF
--- a/ast/src/builder.rs
+++ b/ast/src/builder.rs
@@ -433,7 +433,9 @@ impl Repo {
     }
     fn prepare_file_data(&self, path: &str, code: &str) -> NodeData {
         let mut file_data = NodeData::in_file(path);
-        file_data.name = path.to_string();
+        let filename = path.split('/').last().unwrap_or(path);
+        println!("File name {}", &filename);
+        file_data.name = filename.to_string();
         let skip_file_content = std::env::var("DEV_SKIP_FILE_CONTENT").is_ok();
         if !skip_file_content {
             file_data.body = code.to_string();
--- a/ast/src/builder.rs
+++ b/ast/src/builder.rs
@@ -209,7 +209,6 @@ impl Repo {
             //graph.add_classes(classes);
         }
         info!("=> got {} classes", i);
-        //TODO: add node inheritances to the graph trait
         graph.class_inherits();
         graph.class_includes();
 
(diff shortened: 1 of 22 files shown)


fayekelmith:

I'm unsure why the E2E tests are failing.

ast/src/builder.rs
```diff
@@ -64,9 +78,47 @@ impl Repo {
 
         let mut i = dirs_not_empty.len();
         info!("adding {} dirs... {:?}", i, dirs_not_empty);
+        let mut processed_dirs = HashSet::new();
+
         for dir in &dirs_not_empty {
-            let dir = dir.display().to_string();
-            graph.add_directory(&dir);
+            let dir_path = dir.display().to_string();
```
fayekelmith, line 84: This code block for adding directories is responsible for the alteration of test files. This approach tracks intermediate directories that have an edge: `Directory -> Directory` as in the case of Svelte with the nested directories: 

```bash
│   └── routes
│       ├── api
│       │   └── people
│       │       └── +server.js
```
It picks up `api` as a node and maps the relationship down to the file `+server.js` as edges; hence the increase in nodes and edges count for the Kotlin and Svelte tests modules. Kotlin is particularly heavy with this architecture.

ast/src/builder.rs
```diff
@@ -79,7 +131,30 @@ impl Repo {
...
-            graph.add_file(&filename.display().to_string(), &code);
+
+            let path = filename.display().to_string();
+
+            if graph.find_nodes_by_name(NodeType::File, &path).len() > 0 {
+                // file already exists
+                continue;
+            }
+            //Do not add pkg files here they will be added later
+            if path.ends_with(self.lang.kind.pkg_file()) {
```
fayekelmith, line 142: This restriction is responsible for the change in `ast/src/testing/go/mod.rs` It prevents builder from selecting package files because there is a seperate algorithm to pick them up. It follows this one

ast/src/builder.rs
fayekelmith: Loom video better explaining it:
https://www.loom.com/share/4c72b15172074ca489b54a56a55208e1?sid=b4a1e83e-56ee-476d-9a8a-1f70decae59e

PR #51, opened 2025-04-04
chore: run rust test

tobi-bams:


Evanfeenstra:
Can you change the names
- Run Rust Test on Stakgraph shouild be `Run unit tests`
- the other bigger test should read `Run e2e tests`

PR #70, opened 2025-04-10
Make remaining code sections generic

fayekelmith:
-[X] Make front-end and backend test generic
-[ ] Make individual test generic
-[ ] Make Linker Generic
-[ ] Test BtreeMapGraph


Evanfeenstra:
Any time we can avoid making another trait method, we should. Like why do we need `find_data_model_nodes`? Can't we pass in NodeType::Datamodel?

Evanfeenstra:
Also there is a lot of duplicated code between the ArrayGraph and BTreeGraph in these new functions. Can we share code between them better?

Issue #54, opened 2025-04-07
Svelte Data Models needs some improvements


fayekelmith:
The svelte data model query seems not to be very effective. 

This is a snippet of the Data Models output for 
```rust
 for n in &graph.nodes {
        println!("{:?} {:?}\n\n", n.node_data.name, n.node_type);
    }
```

![Image](https://github.com/user-attachments/assets/eb9165fb-ca58-4d99-b2c4-331bcbd04b4a)

fayekelmith:
Not Just the Data Model is inaccurately queried but other nodes like `CLASS`, `REQUEST`

<img width="1402" alt="Image" src="https://github.com/user-attachments/assets/43d4067e-da78-42a1-aa45-16e4aa582281" />

PR #72, opened 2025-04-11
Svelte queries

turizspace:
Changed queries to use tree-sitter-typescript grammar from tree-sitter-svelte_ng and address #54


Evanfeenstra:
check out this example @turizspace https://github.com/stakwork/stakgraph/blob/main/ast/src/lang/queries/python.rs#L16

You can choose which underlying language to parse with. So you can check the node.file and see the extension. If it ends with .svelte use svelte parser, otherwise typescript

turizspace:
queries modified to use typesitter-svelte-ng only as including typescript queries with the svelte would break syntax. now the data models and classes are captured better than before

PR #73, opened 2025-04-11
Feature/close generics

fayekelmith:
- [X] Made print_json generic
- [x] Single deciding point for graph

### DIFF
--- a/ast/src/builder.rs
+++ b/ast/src/builder.rs
@@ -1,6 +1,5 @@
 use super::repo::{check_revs_files, Repo};
 use crate::lang::graphs::Graph;
-use crate::lang::ArrayGraph;
 use crate::lang::{asg::NodeData, graphs::NodeType};
 use anyhow::{Ok, Result};
 use git_url_parse::GitUrl;
@@ -13,10 +12,7 @@ use tracing::{debug, info};
 const MAX_FILE_SIZE: u64 = 100_000; // 100kb max file size
 
 impl Repo {
-    pub async fn build_graph(&self) -> Result<ArrayGraph> {
-        self.build_graph_inner().await
-    }
-    pub async fn build_graph_inner<G: Graph>(&self) -> Result<G> {
+    pub async fn build_graph<G: Graph>(&self) -> Result<G> {
         let mut graph = G::new();
 
         println!("Root: {:?}", self.root);
--- a/ast/src/builder.rs
+++ b/ast/src/builder.rs
@@ -1,6 +1,7 @@
 use super::repo::{check_revs_files, Repo};
 use crate::lang::graphs::Graph;
 use crate::lang::{asg::NodeData, graphs::NodeType};
+use crate::lang::{ArrayGraph, BTreeMapGraph};
 use anyhow::{Ok, Result};
 use git_url_parse::GitUrl;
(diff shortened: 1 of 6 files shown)


fayekelmith:
@Evanfeenstra, If this implementation of `print_json` is okay, I suggest we have a single point of deciding which graph to implement. Perhaps an ENV like `GRAPH_TYPE` ?

Evanfeenstra:
yes, lets add a ENV var

fayekelmith:
> Wow, I've never seen `downcast_ref` before. 
> 
> ```
> 
> fn as_array_graph<G: Graph + Serialize + 'static>(graph: &G) -> Option<&ArrayGraph> {
> 
>     if let Some(array_graph) = (graph as &dyn Any).downcast_ref::<ArrayGraph>() {
> 
>         Some(array_graph)
> 
>     } else {
> 
>         None
> 
>     }
> 
> }
> 
> ```
> 
> Are you sure you need to manually return Some or None here? I mean, if `downcast_ref` returns an Option already why do u need to match it?

True sir... The match is unnecessary. I'll get rid of it.

fayekelmith:
@Evanfeenstra, I’m a little confused about how to structure the code so that it uses the same Graph type everywhere.

Because we have `build_graph` in so many places, it means if I make `build_graph` generic (which I need to, if the graph type has to be initialized in `index.rs`), then I have to decide which Graph type to continue with, wherever `build_graph` is called. 

Please how would you recommend I handle this?

fayekelmith:
Thinking about it again, it won't be possible to have the Graph Type dynamically inferred at runtime from `ENV`, else we would have to use trait objects and not generics. 

If it has to remain static, then we can't use `ENV`, however, we can simplify things by creating extra methods in `repo.rs` and `builder.rs` as such:

```rust
impl Repo{
pub async fn build_graph(&self)->Result<ArrayGraph>{
self.build_graph_with_type::<ArrayGraph>().await
}
 pub async fn build_graph_with_type<G: Graph>(&self) -> Result<G> {
        let mut graph = G::new();
       //rest of implementation
} 
}

```

With these two methods (assuming same for `build_graphs` in `repo.rs`, we can have it all statically typed but with two points to change the Graph Type.

Evanfeenstra:
Yes exactly. Try not to change `build_graph` everywhere. Maybe we add `build_graph_btree` as a second option

fayekelmith:
@Evanfeenstra, Ok.  I've done that.
Please review the changes.

Evanfeenstra:
Great! Is there a simple test we can add to make sure the ArrayGraph and the BTreeGraph have the exact same output? Like parsing the code in `ast/src/testing/react` with both, and comparing at the end?

fayekelmith:
To test both, we'll have to build both graphs first, and we currently don't. Only one graph is built at a time. 

If we build the graph with one Graph Type and test it with both, only one test will pass.

Evanfeenstra:
> To test both, we'll have to build both graphs first, and we currently don't. Only one graph is built at a time.
> 
> If we build the graph with one Graph Type and test it with both, only one test will pass.

I mean a higher pevel test that parses the react demo code, once with each Graph type. And then saves each one with the print_json thing, and then compared the two outputs

fayekelmith:
Okay! That's possible. 
I'll try with React first.

PR #78, opened 2025-04-15
Graph Testing

fayekelmith:
- [X] Setup testing


fayekelmith:
@Evanfeenstra, Sir, what do you think about this approach? 

What should I improve before adding the other test metric for the other graphs?

fayekelmith:
Also, I notice that in the method `filter_out_nodes_without_children`, ArrayGraph removes the node but does not remove associated edges. 

So, in BTreeMapGraph, where it removes both, there will be a discrepancy in the edge count. 
The particular edge in question here is
`File - db.go` to `Class - Person` 
 
 So, @Evanfeenstra, do I remove the edge in both graphs and then update the testing parameters, or remove just the node (leaving the unreliable edge behind)?
 
 This should only affect the edge counts of the Rust and Go test files.
 
 PS: The test currently passes because I deactivated BTreeMapGraph's ability to remove the edge of the associated node

ast/src/testing/kotlin/mod.rs
```diff
@@ -17,9 +17,11 @@ pub async fn test_kotlin_generic<G: Graph>() -> Result<(), anyhow::Error> {
 
     let graph = repo.build_graph_inner::<G>().await?;
 
+    graph.analysis();
+
     let (num_nodes, num_edges) = graph.get_graph_size();
     assert_eq!(num_nodes, 145, "Expected 145 nodes");
-    assert_eq!(num_edges, 148, "Expected 148 edges");
+    assert_eq!(num_edges, 76, "Expected 148 edges");
```
fayekelmith, line 24: Many edges are being repeated because the Library and Import Nodes are not properly queried, hence it created repeated edges. 



Here is a screenshot of the repetitive nodes and edges. 

Screenshot 2025-04-21 at 3 09 32 PM
Screenshot 2025-04-21 at 3 07 35 PM

ast/src/lang/graphs/btreemap_graph.rs
```diff
@@ -699,41 +722,43 @@ impl Graph for BTreeMapGraph {
...
             .take_while(|(k, _)| k.starts_with(&prefix))
+            .find(|(_, node)| {
+                node.node_data.file == node_ref.node_data.file
+                    && node.node_data.name == node_ref.node_data.name
+            })
             .map(|(_, node)| node.clone())
-            .next()
     }
 
     fn find_node_from_node_key(&self, node_type: NodeType, node_key: NodeKeys) -> Option<Node> {
```
Evanfeenstra, line 746: why does this not work with the ({}-{}-{}, type, name, file)? that will be so much more efficient

ast/src/testing/mod.rs
```diff
@@ -15,6 +18,78 @@ pub mod test_backend;
 pub mod test_frontend;
 pub mod utils;
 
+static TEST_EXPECTATIONS: &[GraphTestExpectations] = &[
```
Evanfeenstra, line 21: this is great! But you can simplify with `..Default::default()` in the struct if you dont needs the lsp or critical edges fields...

ast/src/utils.rs
```diff
@@ -93,3 +93,24 @@ pub fn get_use_lsp() -> bool {
     }
     false
 }
+
+pub fn create_synthetic_key_from_ref(node_ref: NodeRef, start: u32) -> String {
```
Evanfeenstra, line 97: you forgot the verb here 

https://github.com/stakwork/stakgraph/blob/main/mcp/src/graph/utils.ts#L76

Evanfeenstra (changes requested):
just a few small nits

ast/src/testing/mod.rs
fayekelmith: okay! I'll do just that.

Evanfeenstra:
can you explain the react/typescript import_query change? 

There should be one import Node per file, with the exact text that the file has (just the imports at the top of the file)

fayekelmith:
> can you explain the react/typescript import_query change?
> 
> There should be one import Node per file, with the exact text that the file has (just the imports at the top of the file)

Screenshot 2025-04-23 at 11 13 09 AM

current import query

--- 

I noticed that the previous import query got the entire importation line (which is not bad) and not the `imports`. So I thought it would be more useful if the actual imports were collected, combined, and added as a unit per file. 

There is a method `collect_imports` that combines all the imports per file and uniquely identifies them. 

The only difference between this query and the previous is the content of the `imports`.

Evanfeenstra:
No we need the whole line. Remember the point is to show LLM real code snippets. The LLM does a "search" and "replace" of file content, if it needs to add more imports. So it has to be an EXACT match of the real text in the file

fayekelmith:
> No we need the whole line. Remember the point is to show LLM real code snippets. The LLM does a "search" and "replace" of file content, if it needs to add more imports. So it has to be an EXACT match of the real text in the file

Oooh ooh... okay.. I'll revert them.

ast/src/builder.rs
```diff
@@ -503,6 +503,15 @@ pub fn combine_imports(nodes: Vec<NodeData>) -> Vec<NodeData> {
...
         return Vec::new();
     }
+
+    let file_path = &nodes[0].file;
+    let file_name = file_path.split('/').last().unwrap_or(file_path);
+
+    let import_name = if nodes.len() == 1 {
+        nodes[0].name.clone()
+    } else {
+        format!("imports-{}-{}", file_name, nodes.len())
```
Evanfeenstra, line 513: What if two files have the same name in different directories? Just use `create_synthetic_key_from_ref` or some other unique key here

Evanfeenstra:
main:

Screenshot 2025-04-23 at 12 29 22 PM

your branch:
Screenshot 2025-04-23 at 12 31 20 PM

Evanfeenstra:
> your branch:
> 
> Screenshot 2025-04-23 at 12 29 22 PM
> main:
> 
> Screenshot 2025-04-23 at 12 31 20 PM
> any idea why there are so many more imports?

oops my bad, its all good

PR #91, opened 2025-04-23
Fix: Adding remaining languages to the test

fayekelmith:
- [X] Added the rest of the programming language to `test_graph_similarity` test
- [X] Improved `add_edge` to ensure uniqueness by adding `start` to `NodeKey`

### DIFF
--- a/ast/src/builder.rs
+++ b/ast/src/builder.rs
@@ -504,7 +504,7 @@ pub fn combine_imports(nodes: Vec<NodeData>) -> Vec<NodeData> {
     if nodes.is_empty() {
         return Vec::new();
     }
-    let import_name = create_node_key(Node::new(NodeType::Import, nodes[0].clone()));
+    let import_name = create_node_key(&Node::new(NodeType::Import, nodes[0].clone()));
     let mut combined_body = String::new();
     let mut current_position = nodes[0].start;
     for (i, node) in nodes.iter().enumerate() {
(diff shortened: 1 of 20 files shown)


fayekelmith:
The old way of adding edges was inefficient. 

### For ArrayGraph

To ensure non-duplication, `types`, `names`, `files`,  `verbs`, and for function calls `operand` were used to compare the incoming edge with the existing edges by **looping through the list of nodes**, however, some edges fell through the cracks since a more crucial key field `start` was absent.

### For BTreeMapGraph
We used the `find_node_from_node_ref` method, and since `NodeKeys` didn't have the `start` field, a default of 0 was used, which still left some edges unaccounted for. 

### Approach used in this PR
- I added `start` to the NodeKeys to uniquely identify every Edge
- Using all 5 elements that make the unique key (type, name, file, start, verb), we can uniquely identify both edges and nodes (and they match). 

- For ArrayGraph, I created a `create_edge_key` using the `source_key`, `dst_key` and **detailed** `EdgeType` [to handle the case for Calls(meta)]
- create_edge_key uniquely identifies all edges now and prevents de-duplication while leaving no connection behind. 

- For BTreeGraph, I used `create_node_key_from_ref` to create correct keys for the edge then, they can be directly added to the Graph. 

--- 
Reasons for altering tests:
- By adding `start` to `NodeKeys`, some edges that weren't previously uniquely identified because their IDs were faulty. It helped make the change from: 
`type-name-file-0-verb` --> `type-name-file-start-verb` 
which is more accurate, hence picking up new non-repeating edges. 
--- 
Optimization: There is no need to query the nodes before adding edges; keys identical to the node_keys can be directly generated and used to ensure the uniqueness of each node.

fayekelmith:

PR #96, opened 2025-04-25
Add Tests for nodes and edges to the Test Graph Similarity test

fayekelmith:
- [X] Added React
- [X] Added Go 
- [x] Add TypeScript 
- [x] Add Python 
- [x] Add Rust 
- [x] Add Java 
- [x] Add Angular
- [x] Add Svelte
- [x] Add Ruby
- [x] Add Swift
- [x] Add Kotlin


ast/src/testing/graphs.rs
```diff
@@ -1,19 +1,43 @@
...
-    pub _critical_edges_lsp: Option<Vec<CriticalEdgeCheck>>,
+    pub expected_lsp_nodes: Option<u32>,
+    pub expected_lsp_edges: Option<u32>,
+    pub edges: Vec<EdgeCheck>,
+    pub lsp_edges: Option<Vec<EdgeCheck>>,
+    pub nodes: Vec<NodeCheck>,
+    pub lsp_nodes: Option<Vec<NodeCheck>>,
 }
 
 impl Default for GraphTestExpectations {
```
Evanfeenstra, line 43: why are you defining the Default, can you just use the proc macro `#[derive(Default)]` ?

ast/src/testing/graphs.rs
```diff
@@ -61,65 +87,278 @@ pub async fn run_graph_similarity_test(
...
+                    target_name,
+                    edge_type,
+                    expectations.lang_id
+                );
+            }
+        }
+    }
+
     Ok(())
 }
```
Evanfeenstra, line 364: wow 1000 lines of code is alot to maintain... i was hoping for every language to pass the exact same tests. Because we have the mod.rs tests in each separate language anyways

Im not sure what the best approach is

ast/src/testing/graphs.rs
fayekelmith: Okay... I'll do that sir.

ast/src/testing/graphs.rs
fayekelmith: There are some fundamental test common to most languages... I can abstract them. 

The length is because this test is a replica of the other modular tests...

I tested the same aspects - and a little more... I can remove them and let them reside originally where they live except languages that don't have the modular unit tests.

fayekelmith: There are some fundamental test common to most languages... I can abstract them. 

The length is because this test is a replica of the other modular tests...

I tested the same aspects - and a little more... I can remove them and let them reside originally where they live except languages that don't have the modular unit tests.

PR #100, opened 2025-04-30
Neo4j Graph Setup

fayekelmith:
- [X] Neo4j graph setup with Docker container.


fayekelmith:
@Evanfeenstra, this is the foundation for the Neo4j graph. Is there something else you would prefer?

Evanfeenstra:
- there is already a setup in `mcp` for docker neo4j. Just use that
- put the entire neo4j graph (and neo4rs dep) behind a cargo feature https://doc.rust-lang.org/cargo/reference/features.html

ast/src/testing/mod.rs
```diff
@@ -8,6 +8,7 @@ pub mod angular;
 pub mod go;
 pub mod graphs;
 pub mod kotlin;
+pub mod neo4j_graph;
```
Evanfeenstra, line 11: put the `#[cfg(feature = "neo4j")]` here instead of in the neo4j_graph.rs file

PR #102, opened 2025-04-30
Implement the Graph Trait for Neo4jGraph

fayekelmith:
- [x] create helper methods
- [x] use helper methods to implement Graph Trait for Neo4jGraph
- [x] Write tests to ensure it works.

### DIFF
--- a/ast/src/builder.rs
+++ b/ast/src/builder.rs
@@ -27,21 +27,15 @@ impl Repo {
             use crate::lang::graphs::{
                 neo4j_graph::Neo4jGraph, neo4j_utils::Neo4jConnectionManager,
             };
-            use std::any::TypeId;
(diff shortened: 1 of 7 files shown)


fayekelmith:
@Evanfeenstra, I focused this PR on implementing the Graph Trait. 

Queries and methods of the Graph trait are not all correct. In subsequent PRs I will fix that. But, what do you think about this approach, sir?

ast/src/builder.rs
```diff
@@ -22,6 +22,23 @@ impl Repo {
         self.build_graph_inner().await
     }
     pub async fn build_graph_inner<G: Graph>(&self) -> Result<G> {
+        #[cfg(feature = "neo4j")]
+        {
```
Evanfeenstra, line 26: to keep this function clean, can you move this code into a separate util function

PR #103, opened 2025-05-02
fix ruby class queries

fayekelmith:
Fix Ruby class queries.

### DIFF
--- a/ast/src/lang/queries/ruby.rs
+++ b/ast/src/lang/queries/ruby.rs
@@ -40,31 +40,34 @@ impl Stack for Ruby {
             ) @{LIBRARY}"#
         ))
     }
+
     fn class_definition_query(&self) -> String {
         format!(
-            r#"[
+            r#"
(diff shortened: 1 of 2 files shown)


ast/src/lang/queries/ruby.rs
```diff
@@ -40,31 +40,34 @@ impl Stack for Ruby {
...
+            (
                 (class
-                    name: [
-                        (constant)
-                        (scope_resolution)
-                    ] @{CLASS_NAME}
-                    (superclass
-                        (constant) @{CLASS_PARENT}
-                    )?
+                    name: (_) @c{CLASS_NAME}
```
Evanfeenstra, line 49: what is this `@c{CLASS_NAME}` @fayekelmith

fayekelmith: ooh, typo... it should be @{CLASS_NAME}. fixing it.

PR #105, opened 2025-05-05
chore: add var nodetype

tobi-bams:


Evanfeenstra:
VARIABLE_DECLARATION should be the `body`, and IDENTIFIER should be the `name` (but call it VARIABLE_NAME please) @tobi-bams

ast/src/lang/parse.rs
```diff
@@ -105,6 +106,28 @@ impl Lang {
...
+        Self::loop_captures_multi(q, &m, code, |body, node, o| {
+            let mut impy = NodeData::in_file(file);
+            if o == IDENTIFIER {
+                impy.name = body.to_string();
+                impy.body = body;
+                impy.start = node.start_position().row;
+                impy.end = node.end_position().row;
+                res.push(impy);
+            }
+
```
Evanfeenstra, line 126: Here @tobi-bams the VARIABLE_DECLARATION should be set to be the body

PR #107, opened 2025-05-06
Fix/react var

tobi-bams:


Evanfeenstra:
cool idea.

But `is_top_level` will only work for ts, other ones might be different

What if you remove the 
```
"function_declaration"
                    | "class_declaration"
                    | "method_definition"
                    | "arrow_function"
                    ```
                    
 and only check for `program`.. .will that work? And then for each language add a "program_name" method to the  `Stack` trait. You can make it return the string "program" by default. But if we find that other languages DONT use "program" for the top level, we can add special ones for them
 
 does that make sense @tobi-bams

**Comment by tobi-bams** [comment]
@Evanfeenstra  Removing it will be an issue because we need to check if it's a child of a function and stop it
```
"function_declaration"
                    | "class_declaration"
                    | "method_definition"
                    | "arrow_function"
                    ```
```

tobi-bams:
> oh i get it now.
> 
> But... what do you think? Do you even need to `loop` over the parent, or can you just check the immediate parent?

Yea, this is a valid point but for unique cases like` export`, variables are not direct children of program we have to loop to ensure they are.

Evanfeenstra:
I think the `Self::is_top_level` should only apply to the VARIABLE_DECLARATION node. Can you change that and test it to make sure it captures all the top level vars in a file?

PR #115, opened 2025-05-09
Fix failed graph edges addition

fayekelmith:


fayekelmith:
@Evanfeenstra, please I'm a little stuck and perhaps you could help me point out where I'm missing the point. 
In the graph, edges in the `add_calls` method are been created (and logged) but unfortunately not seeded in the database. The queries execute successfully (as logged in the terminal) but when I query them, they don't seem to be there. 

Please, am I missing something? 

Test : `cargo test --features neo4j test_neo4j -- --ignored` (after starting neo4j db using docker in `/mcp`)

PR #121, opened 2025-05-15
Feature: Updating Graph DB on commit

fayekelmith:
- [X] Setup `process` endpoint
- [X] Set up a GitHub workflow to trigger the `process` endpoint
- [X] Test with new commits to ensure it works (for demo repo)


fayekelmith:
@Evanfeenstra, please review this work. 

I wrote a Github Workflow for the demo-repo here: https://github.com/stakwork/demo-repo/pull/3
It requires configuration to get the Variable Environment variables for the GRAPH_SERVER however, I have tested it locally and it works. It's able to trigger the endpoint to rebuild the graph.

fayekelmith:
@Evanfeenstra, I think this approach will likely create an inaccurate graph. 

If I understood the approach clearly, it is supposed to identify the modified files, build a subgraph out of them, and then extend the original graph with the subgraph. 

However, I think this approach assumes that a resultant graph can be obtained by exclusively summing up graphs which is inaccurate. 

In the case of function calls, a modification in file A might establish a `CALLS` relationship in unmodified file B. If we sum up the graphs, we'll not have that relationship. And there is much more like this in production-grade code. 


A potential solution would be to collect both the modified graph and its dependencies, which introduces a new challenge. 

### How do we identify files with dependencies using AST?

I think LSP is ideal for this, but it's slower. Trade-off.

What do you think about it, @Evanfeenstra?

fayekelmith:
@Evanfeenstra, I've preserved the incoming edges and used them to link both graphs.

Evanfeenstra:
@fayekelmith 

- so you tested this with the full flow of uploading from `jsonl` files using the ts server, and then hooking up to the same neo4j to update the graph?
- what about testing with REPO_PATH instead of REPO_URL? So you can work on your local repo and hit `process` when you commit locally
- I dont fully understand the `process` logic. Does `deleted_files` have all the same files as `changed_files`? Is there any way to simplify or modularize this more?

There should be strong tests for this process. like
- file A removes edge to file B. Is that removed when file A is re-processed? Are existing edges the other way around (B to A) preserved when B is unchanged?
- please write a ts test with simple functions that call each other and test re-processing the graph

fayekelmith:
Alright, @Evanfeenstra, I'll write some tests with a demo repo. However, this is what `/process` accomplishes:

- Gets repo URL from the request
- Then pulls the repo locally to `/tmp/repo-name` to get the latest changes, or if the repo is new, it clones the repo.
- Get the current repo's hash code `current_hash`
- Get the hash code from the Repo in the graph `stored_hash`
- Using the hashes, I compare to get the changed files. 
- I sort out the files that were deleted in the latest commit and remove them from the Graph [looking at it now, I don't need to handle this separately :) ]
- Then I move to the modified files and collect all the edges *that have a target in the changed file but do not have a source in the same file*
- Remove the nodes from the graph that belong to that file
- Build a subgraph with just that file [I'll update it to process all at once]
- Extend the original graph with the subgraph
- Lastly, I loop through the list of edges I saved earlier and re-establish the relationship if the source and target still exist. Hence, we don't lose the glue between modified and unmodified files. 

For new repos, it's a little straightforward
---

Now that I've explained, I have noticed blind spots. I'll: 
- Remove the deleting logic and maintain just modification
- Upgrade to bulk processing
- Add tests to demonstrate Graph updates. 

@Evanfeenstra, Is there anything else I might be missing?

Evanfeenstra:
Sounds good. Take time with it, to make it as simple as possible! Its pretty confusing since there are many kinds of edges
- modified file to unmodified file (these need to be captured)
- unmodified file to modified file (these need to be saved and re-applied. they will not have changed)
- modified file to modified file (just make sure these are properly added, since you are making a subgraph per file?)

The real logic should probably like in a new mod in `ast`, where it can be thoroughly tested. The `standalone` crate should be minimal (just the axum stuff). We'll add full bulk graph upload here for new graphs, and maybe more endpoints too

fayekelmith:
> * modified file to unmodified file (these need to be captured)
> * unmodified file to modified file (these need to be saved and re-applied. they will not have changed)
> * modified file to modified file (just make sure these are properly added, since you are making a subgraph per file?)

Okay, @Evanfeenstra 

I have done just that. 

- Refactored the code to make standalone lean and handle only server work
- I moved the processing logic to a mod in `ast`, including the testing (alongside other tests)


For the repository, I did something a little different. 
I simulated **all three update scenarios—modification, addition, and deletion—by using a static demo repository and two carefully crafted commits.** This allowed me to control the relationships between functions across files and observe how the graph changes in response to each type of update.

Specifically:

- **Modification**: In the second commit, I modified one file (e.g., alpha.go) to change which functions it calls, simulating a change in relationships from a modified file to both modified and unmodified files.
- **Addition**: I added a new file (delta.go) with a function that introduces new edges to the graph, testing how new relationships are incorporated.
- **Deletion**: I deleted an existing file (beta.go), ensuring that all edges to and from its functions are properly removed from the graph.

By running the test across these two commits, I can assert:

- Edges from modified to unmodified files are updated or removed as expected.
- Edges from unmodified to modified files are preserved if the unmodified file remains.
- Edges involving deleted files are fully removed.
- New edges from added files are correctly created.

This approach gives us a deterministic and repeatable way to verify the correctness of incremental graph updates for all key scenarios, without relying on dynamic or external repo changes.


For trial purposes, I used this demo repo: https://github.com/fayekelmith/demorepo
If you agree with the approach, I'll make similar changes to the `stakgraph/demo-repo` so we can use it for testing.

Evanfeenstra:
Great! We should also make a github action to publish the `standalone` as a docker image to ghcr.io/stakwork/stakgraph. There is already a docker file at Dockerfile.standalone (I haven't tested it)

Evanfeenstra:
try to avoid `rustlang/rust:nightly` in the docker image, because it can change in the future and break something. Better to stick to a stable version @fayekelmith

fayekelmith:
> try to avoid `rustlang/rust:nightly` in the docker image, because it can change in the future and break something. Better to stick to a stable version @fayekelmith

alright. I'll fix that.

Issue #120, opened 2025-05-15
import node_key too long


Evanfeenstra:
some import node_key are way too long, like for Go where it puts together the whole body of the import. lets simplify ... node_key should be the same for all node types @fayekelmith

fayekelmith:
Okay sir, is this specific only for the node_key of `IMPORTS`? 

Because I think the length issue comes in here : https://github.com/stakwork/stakgraph/blob/main/ast/src/lang/parse.rs#L98

where the name is registered as the body, and then it's used by `create_node_key` to form the lengthy node keys. 

Can I change `impy.name = body.to_string()` to `impy.name = "imports"` this way, it's guaranteed to be short? 

Does this approach work? @Evanfeenstra

Evanfeenstra:
yes that's fine @fayekelmith. it will always be unique since both "file" and "start" are in the key

PR #122, opened 2025-05-15
Import names to constant

fayekelmith:
#120


fayekelmith:
@Evanfeenstra, here is the fix for the import name's length.

Issue #134, opened 2025-05-20
Ruby model dependencies


Evanfeenstra:
Ruby on Rails models can depend on each other

For example this here `has_many :articles, dependent: :destroy`

https://github.com/stakwork/stakgraph/blob/main/ast/src/testing/ruby/app/models/person.rb#L2

So we need to add Class - CALLS -> Class for each of these

To start lets just use simple finding using file names

```
belongs_to
has_one
has_many
has_many :through
has_one :through
has_and_belongs_to_many
```

PR #138, opened 2025-05-20
Adding ruby Class association edge.

fayekelmith:
- [X] Improved classed parsing to take care of Ruby class association
- [x] Updated Ruby queries and tests to collect class association.

Adresses issue : #134

### DIFF
--- a/ast/src/builder.rs
+++ b/ast/src/builder.rs
@@ -244,16 +244,21 @@ impl Repo {
         i = 0;
         info!("=> get_classes...");
         for (filename, code) in &filez {
-            let classes = self.lang.get_classes::<G>(&code, &filename)?;
+            let qo = self
+                .lang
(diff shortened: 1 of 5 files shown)


fayekelmith:
@Evanfeenstra please what do you think about this implementation?

Evanfeenstra:
@fayekelmith i dont see the specific associations in the query
```
belongs_to
has_one
has_many
has_many :through
has_one :through
has_and_belongs_to_many
```
there are others that are NOT about relationships between classes (like `validates`), so I think we need to have those in the query

Evanfeenstra:
also can you add the specific edge test, for example Person - CALLS -> Article

Evanfeenstra:
looks like you deleted some changes i made earlier today with the mailer prefix stuff

fayekelmith:
> looks like you deleted some changes i made earlier today with the mailer prefix stuff

ooh, I didn't notice, let me check it out and revert it.

Issue #139, opened 2025-05-20
remove CallsMeta


Evanfeenstra:
we dont really need CallsMeta, and it gets in the way of finding nodes by edge type.

remove CallsMeta

and this will make it easier to test the changes from https://github.com/stakwork/stakgraph/pull/138

PR #142, opened 2025-05-21
removed CallsMeta and added ruby class call test

fayekelmith:
This PR Attends to #139 

and provides the necessary tests for #134

### DIFF
--- a/ast/src/lang/graphs/array_graph.rs
+++ b/ast/src/lang/graphs/array_graph.rs
@@ -293,13 +293,7 @@ impl Graph for ArrayGraph {
                 self.add_edge(rt);
             }
             for r in reqs {
-                let edge = Edge::calls(
-                    NodeType::Function,
-                    &node,
-                    NodeType::Request,
-                    &r,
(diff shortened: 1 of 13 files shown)


fayekelmith:
@Evanfeenstra, 

In the ruby test file, we have two class `Article` and `⁣Person`
both have relations with one another `Article <- belongs_to <- Person` and `Person -> has_many` Articles. 

With the new implementation of `CALLS` we can't differentiate between both... They establish just one relationship between both classes `Persn - CALLS - Artice`. Is that fine?

Evanfeenstra:
> @Evanfeenstra,
> 
> In the ruby test file, we have two class `Article` and `⁣Person` both have relations with one another `Article <- belongs_to <- Person` and `Person -> has_many` Articles.
> 
> With the new implementation of `CALLS` we can't differentiate between both... They establish just one relationship between both classes `Persn - CALLS - Artice`. Is that fine?

Yes thats fine. Just as long as the relationship is identified

PR #145, opened 2025-05-22
Update Standalone Graph

fayekelmith:
- [X] Updated flow to use ENV
- [X] Added an `injest` endpoint

### DIFF
--- a/standalone/src/handlers.rs
+++ b/standalone/src/handlers.rs
@@ -1,33 +1,58 @@
-use crate::types::ProcessBody;
 use crate::types::{AppError, ProcessResponse, Result};
 #[cfg(feature = "neo4j")]
 use ast::lang::graphs::graph_ops::GraphOps;
-use ast::lang::Graph;
 use ast::repo::Repo;
 use axum::Json;
-use lsp::git::{get_commit_hash, git_pull_or_clone};
-use tracing::info;
-pub async fn process(Json(body): Json<ProcessBody>) -> Result<Json<ProcessResponse>> {
-    let repo_url = body.repo_url.clone();
-    let username = body.username.clone();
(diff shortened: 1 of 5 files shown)


fayekelmith:

standalone/src/handlers.rs
```diff
@@ -109,3 +137,94 @@ pub async fn clear_graph() -> Result<Json<ProcessResponse>> {
...
+        let url = repo_url.unwrap();
+        let tmp_path = Repo::get_path_from_url(&url)?;
+        (tmp_path, url, true)
+    };
+
+    let username_clone = username.clone();
+    let pat_clone = pat.clone();
+
+    let result = tokio::task::spawn_blocking(move || {
+        #[cfg(feature = "neo4j")]
```
Evanfeenstra, line 165: i dont think you need to check for neo4j feature here, this `standalone` bin must have neo4j enabled. you can enable it by default in the standalone Cargo.toml

standalone/src/handlers.rs
```diff
@@ -1,34 +1,59 @@
...
 use ast::repo::Repo;
 use axum::Json;
-use lsp::git::{get_commit_hash, git_pull_or_clone};
-use tracing::info;
-pub async fn process(Json(body): Json<ProcessBody>) -> Result<Json<ProcessResponse>> {
-    let repo_url = body.repo_url.clone();
-    let username = body.username.clone();
-    let pat = body.pat.clone();
+pub async fn process() -> Result<Json<ProcessResponse>> {
+    let repo_path = env_not_empty("REPO_PATH");
```
Evanfeenstra, line 7: I actually like the repo_url and repo_path being in the body instead. Why not support both? they can exist in the env, but be overwritten by values in the body.

PR #146, opened 2025-05-22
Fix : Fortify Neo4jGraph

fayekelmith:
- [X] Python test
- [ ] Kotlin test
- [ ] Swift test
- [ ] Angular and Svelte test
- [ ] Update ArrayGraph and BTreemap edge creation to match only node definition, not appellation. 
- [ ]  Compare Graphs Test


ast/src/lang/graphs/neo4j_utils.rs
```diff
@@ -206,27 +210,27 @@ impl EdgeQueryBuilder {
...
         let rel_type = self.edge.edge.to_string();
-        
         let source_type = self.edge.source.node_type.to_string();
         let target_type = self.edge.target.node_type.to_string();
-        
-        // this is ideal query for unique edges
+
+        //ideal query
         // let query = format!(
         //     "MATCH (source:{} {{name: $source_name, file: $source_file, start: $source_start}}), \
```
fayekelmith, line 221: @Evanfeenstra 

This query would be ideal to pick up nodes from the graph before establishing a connection, but when I use it, it loses so many edges, instead. The query without the `start` field works. It only fails when we have multiple nodes of the same type and same name in the same file but at different locations. This case comes up in `Swift`. 

The changes I made to ArrayGraph and BtreeMapGraph are to make sure that a connection is established only between node definitions, not calls. Which corresponds to the logic that Neo4j uses to establish an edge between nodes (i.e the non-ideal query). 

However, because of the edge case in Swift, I tried implementing adding the `start` field again, but I just can't get it to work. Please @Evanfeenstra , can you check this out and perhaps point out something I am missing out on?

fayekelmith: The changes in the edge count is for the reason I highlighted up there. I updated the ArrayGraph and BTreeMapGraph to establish edges only between FUNCTION/CLASS definition -> FUNCTION definition not Calls which will establish many of them. This is the approach the Neo4jGraph uses because it only establishes relationships between nodes it has in its graph (definitions, not calls.) 

An alternative (a better solution, actually) would be to make the Neo4jGraph use the `start` field for edge creation, which is proving to be really hard to do.

Evanfeenstra: its fine not to hvae multiple edges for each call within a function to the same function

We can later on work on distinguishing edges to functions of the same name in the same file

PR #150, opened 2025-05-26
Getting all react top level and exported Variables

fayekelmith:
- [X] Updated react queries to capture only exported and top level variables.
- [X] Updated tests to match new variables


fayekelmith:

Evanfeenstra:
Great! I guess the `+` and the `[]` were messing it up before

Can we ensure that no "const" function is captured. Like `export const hello = () => {}` should NOT be captured, since that will already be captured in the function query

Issue #149, opened 2025-05-26
IMPORTS edge


Evanfeenstra:
We should have edges for files importing other stuff

for example File - IMPORTS -> Function

Parsing each import to create an edge to each other thing imported into the file

PR #151, opened 2025-05-27
Implementing the Imports Edge

fayekelmith:
This PR addresses issue #149 

- [X] Update react queries to give information about Import names and destination
- [x] Create helper methods to establish edges between `File - IMPORTS -> Function/Class/Datamode/Var`
- [x] Parse them and add trigger `builder.rs`

Subsequent PRs will be to update queries for other languages.


ast/src/lang/parse.rs
```diff
@@ -1212,6 +1271,18 @@ fn _pick_target_file_from_graph<G: Graph>(target_name: &str, graph: &G) -> Optio
 
     target_file
 }
+fn resolve_import_path(import_path: &str) -> String {
```
Evanfeenstra, line 1274: this probably should be in the trait? since different languages handle things differently. Like go imports from the "package" not from a file

ast/src/lang/queries/react.rs
```diff
@@ -36,11 +36,38 @@ impl Stack for ReactTs {
...
+    //     Some(format!(
+    //         r#"(program
+    //             (import_statement)+ @{IMPORTS}
+    //         )"#,
+    //     ))
+    // }
+
     fn imports_query(&self) -> Option<String> {
         Some(format!(
-            r#"(program
```
Evanfeenstra, line 41: this will break the existing Import nodes? Currently, we want ALL the imports in a file to be a single node with multiple lines in the file. Please make sure this still work. If we need two separate queries for the Import nodes and the IMPORTS edge, thats fine

fayekelmith: It doesn't break the existing import nodes... With the current import nodes, the logic to combine all the imports in a file into a single node is in the `builder.rs` file, and as long as every query still maintains the `IMPORTS` constant, everything should remain the same. The new elements of the query are only used in an isolated method to construct and collect the import edges.

fayekelmith:
@Evanfeenstra, please, when you can, check out #150 first; I think after merging it, there will be some conflicts with this PR, and secondly, I would like to try importing `Var` elements and test if it works for them too.

ast/src/lang/queries/react.rs
Evanfeenstra: can you write a test for it? collecting multiple lines of imports in one node. 

The query before had a `+` in it, so it does seem like a different behavior

Evanfeenstra:
> @Evanfeenstra, please, when you can, check out #150 first; I think after merging it, there will be some conflicts with this PR, and secondly, I would like to try importing `Var` elements and test if it works for them too.

ok, merged #150

ast/src/lang/queries/react.rs
fayekelmith: @Evanfeenstra, I've added the test sir. 

Also, I logged the imports: 

Screenshot 2025-05-28 at 12 41 11 AM

If you look at the import nodes, you'll see that in the body, each of them has all the import lines of the file combined in them.

Evanfeenstra:
@fayekelmith I ran the react test and logged the bodies of the captured import sections. for example here is one:
```
===Import: import-imports-srccomponentpatternstsx-0===
import React, { useState, useEffect } from 'react';
import React, { useState, useEffect } from 'react';
```

Evanfeenstra:
And here's another:
```
===Import: import-imports-srccomponentsnewpersontsx-0===
import { useState } from 'react';
import styled from 'styled-components';
import * as api from '../api';
```

Perhaps the two named imports in the first one is causing two captures...

PR #154, opened 2025-05-28
feat: extract RENDERS edges for Rails partials from view files

jordan-ae:
This PR adds logic to extract RENDERS edges between view files and partials in a Rails application. It supports parsing render calls in .rb, .erb, and .haml files and maps them to their corresponding partial files.

### DIFF
--- a/ast/src/builder.rs
+++ b/ast/src/builder.rs
@@ -162,6 +162,14 @@ impl Repo {
             };
 
             graph.add_node_with_parent(NodeType::File, file_data, parent_type, &parent_file);
+
+            if self.lang.use_partial_finder() {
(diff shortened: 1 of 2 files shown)


Evanfeenstra:
Not sure I understand the approach. Just removing the `_` and finding a parent with the same name doesn't solve the issue. We need to create partials in the ruby test like they would be created in a real ruby on rails app

For example one haml file might have 
```
  .table-contents-container
    = render partial: 'control_jobs_table', locals: { skill_maps: @skill_maps }
 ```
 
 And then there would be a file called `_control_jobs_table.haml` which is a partial
 
 I am not a RoR expert, I'm just copying this from a real repo

PR #156, opened 2025-05-28
Variables and Import Edges

fayekelmith:
- [X] Python variables
- [x] Import Edges
- [X] Typescript Variables and Import Edges
- [X] Typescript Test
- [X] Angular Variables and Import Edges
- [X] Kotlin Variables and Import Edges
- [X] Added Rust Variables and Test
- [X] Added Rust import edge queries and updated test server to confirm it works

### DIFF
--- a/ast/src/lang/parse.rs
+++ b/ast/src/lang/parse.rs
@@ -1132,14 +1132,15 @@ impl Lang {
             if let Some(source_path) = import_source {
                 let resolved_path = self.lang.resolve_import_path(&source_path, file);
 
-                for name in &import_names {
+                for import_name in &import_names {
                     for nt in [
                         NodeType::Function,
                         NodeType::Class,
                         NodeType::DataModel,
                         NodeType::Var,
                     ] {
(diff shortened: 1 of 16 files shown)


Evanfeenstra (commented):
Please can you add specific tests to make sure the actual body of the imports is what's expected. These are very important to get right. They must exactly match the text that's in the file. For example

```ts
import {useState} from 'react';

import * as api from './api.ts';
```

Should preserve the empty space between the two imports. 

Please add actual body text into the tests for imports for these languages

fayekelmith:
> Please can you add specific tests to make sure the actual body of the imports is what's expected. These are very important to get right. They must exactly match the text that's in the file. For example
> 
> ```ts
> import {useState} from 'react';
> 
> import * as api from './api.ts';
> ```
> 
> Should preserve the empty space between the two imports.
> 
> Please add actual body text into the tests for imports for these languages

I've added the specific test of body imports of one file per language, sir. 

@Evanfeenstra

PR #159, opened 2025-05-29
multi platform workflow

fayekelmith:
Updated workflow to support multiple architecture docker file.


fayekelmith:
@Evanfeenstra, this should permit the build to operate on different platforms

Issue #158, opened 2025-05-29
vars edges


Evanfeenstra:
```
​​​
import {rootUrl} from ‘./api.ts’

function callApi() {
  const url = ​${rootUrl}/endpoint​
  fetch(url)
}

```

The `callApi` Function should have a CALLS edge to the `rootUrl` Var

PR #161, opened 2025-05-30
Add `Function` - `CALLS` -> `Var`

fayekelmith:
- [X] Use body scanning to get variables called in the method.
- [x] Implement LSP approach. 
- [x] Try optimizing where possible

Addressing #158

### DIFF
--- a/lsp/src/language.rs
+++ b/lsp/src/language.rs
@@ -137,7 +137,8 @@ impl Language {
         match self {
             Self::Rust => true,
             Self::Go => true,
-            Self::Typescript | Self::React => true,
+            Self::Typescript => false,
+            Self::React => true,
             Self::Python => false,
             Self::Ruby => false,
             Self::Kotlin => false,


fayekelmith:
@Evanfeenstra since not only Vars are called (or imported as referenced in #158 ), sometimes Data models are called in methods, e.g 

```rust
async fn create_person(person: web::Json<Person>) -> impl Responder {
    match Database::new_person(person.into_inner()).await {
        Ok(created_person) => HttpResponse::Created().json(created_person),
        Err(err) => {
            let error_message = err.to_string();
            HttpResponse::InternalServerError().json(json!({ "error": error_message}))
        }
    }
}
```

or 
```rust
    async fn get_person_by_id_impl(&self, id: u32) -> Result<Person> {
        let person: Person =
            sqlx::query_as::<_, Person>("SELECT id, name, email FROM people WHERE id = ?")
                .bind(id as i32)
                .fetch_one(&self.pool)
                .await
                .context("Person not found")?;

        Ok(person)
    }
```

Which I think needs an edge `Function` - Uses -> `DataModel`; I'll implement this in subsequent PRs
This PR only establishes `Function` - CALLS - `Var`

Could you please comment on the approach? Is it viable?

Evanfeenstra:
`get_use_lsp` should be used only for testing (forcing either LSP or no LSP)

While the default_do_lsp should be used for real repo parsing. Don't mix them up

Issue #162, opened 2025-05-30
LSP is unstable with typescript test


fayekelmith:
When using LSP implementations in `ast/src/testing/typescript/mod.rs`, the edge counts are unstable. They keep changing on every test run.

fayekelmith:
#### LSP Edge Count Instability 

After further investigation, the instability in edge counts when running the LSP-enabled TypeScript tests appears to be caused by non-deterministic behavior in the LSP server’s responses to `GotoDefinition` requests.

#### Observation
- The total number of edges in the graph fluctuates between test runs, even with no code changes.
- Specifically, the number of Contains edges `(Function → Var)` created by `collect_var_call_in_function_lsp` is inconsistent.
- For example, in the function `initDatabases` (in `src/index.ts`), sometimes edges like:

```bash
initDatabases (Function@src/index.ts) -> sequelize (Var@src/config.ts)
initDatabases (Function@src/index.ts) -> port (Var@src/index.ts)
```
They are both present, and at other times, only one is present (usually just the `sequelize` edge).


#### Debug Effort:
From debug logs we have that: 

```bash
[VAR_CALL_LSP] Considering edge: initDatabases -> port (Var) [key: src/index.ts:initDatabases:port:src/index.ts:variable]
[VAR_CALL_LSP] ADD EDGE: initDatabases (Function@src/index.ts) -> port (Var@src/index.ts)
```

Affected Edges:
The edges most often affected are `Contains` edges from `initDatabases` in `ast/src/testing/typescript/src/index.ts` to variables such as:

- sequelize `ast/src/testing/typescript/src/config.ts`
- port `ast/src/testing/typescript/src/index.ts`
- AppDataSource `ast/src/testing/typescript/src/config.ts`


More info:
- This instability is observed regardless of graph implementation (`ArrayGraph` or `BTreeMapGraph`).
- The problem is intermittent: sometimes all expected edges are present, sometimes one or more are missing. Currently success rater is about 90%
- The issue does not appear to be related to tree-sitter queries or our graph-building logic, but rather to the timing and state of the LSP server during test execution.

#### Hypothesis
- The LSP server sometimes fails to resolve certain identifiers (like port or AppDataSource) in the function body, returning `None` or an unexpected result for `GotoDefinition`
- This causes the corresponding edge not to be created, leading to a mismatch in the expected edge count.

PR #164, opened 2025-06-02
Fixing unstable Typescript LSP

fayekelmith:
Making LSP consistent for ArrayGraph and BtreeMap  in TS. Addressing #162 
- [X] Narrowed down issue to import_edges
- [x] Figure out non-deterministic source of problem.


Evanfeenstra (commented):
Instead of using println, `use test_log::test`, and standard debug!, info! lines etc

```
#[test(tokio::test(flavor = "multi_thread", worker_threads = 2))]
```

PR #165, opened 2025-06-03
Fix: Speed up `/injest` endpoint in standalone

fayekelmith:
- [ ] Setup workflow to process graph with BTreeMap Graph
- [ ] Upload the BTreeMap to Neo4jGraph.


Evanfeenstra:
@fayekelmith I'm not sure I understand.

How is it possible to use all the methods in parse.rs without neo4j_graph implementing Graph?

Issue #177, opened 2025-06-09
Need a Github Actions for Standalone


fayekelmith:
We need to implement a GitHub action similar to `e2e-test.yml` to make sure nothing breaks in the `standalone` crate.

Issue #175, opened 2025-06-09
Inconsistency in numbers formating for uploaded Graph


fayekelmith:
From the uploaded graph, the formats for `start`, `end`, and `token_count` are inconsistent. They are not numbers.

PR #183, opened 2025-06-10
standalone github actions

fayekelmith:
- [x] Github action for e2e Tests
- [x] Github actions for Standalone Unit tests
- [x] More unit test to make sure nothing breaks
- [ ] Address : #177 #175


fayekelmith:
@Evanfeenstra 
Screenshot 2025-06-11 at 9 45 53 AM


I the test, we have expected output for `start` and `stop` positions of `file: stakwork/demo-repo/main.go` and `file: stakwork/demo-repo/routes.go` as `N/A` 

Given that start and stop are numbers, shouldn't this be incorrect? it's supposed to be 0 right? just like in the nodes, the start and end are 0 for files, directories and repositories. 

or am I missing something?

fayekelmith:
@Evanfeenstra to avoid bloating the PR. I've simplified it by removing the e2e-test.. it has issues that need me cleaning up Neo4jGraph and Utils -- which I will do in the next PR. 

Please review this PR; and if we are good with the work here then we can merge it. I'll continue with it in the next PR.

PR #188, opened 2025-06-10
Feat : Class calls DM

fayekelmith:
- Draft one of action plan

### DIFF
--- a/ast/src/builder.rs
+++ b/ast/src/builder.rs
@@ -311,10 +311,16 @@ impl Repo {
                     &st.file,
                 );
             }
+            for dm in &structs {
+                let edges = self
+                    .lang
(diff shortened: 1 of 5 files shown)


Evanfeenstra:
for ruby you can:
1. get datamodels
2. convert to singular + PascalCase (same as convert_association_to_name)
3. add ​`Controller​` or ​`Blueprint`​ on the end
4. look for Class with that name

Dont forget a test for ruby!

Hmmm.... it looks like Blueprints are singular, but Controllers are plural. Maybe just search twice, once for plural once for singular? That might be simplest and I dont think there is any risk of false positive

Also make it flexible so we can add other things besides `Controller` and `Blueprint` easily later on

PR #220, opened 2025-06-18
update /process endpoint 'standalone'

fayekelmith:
- [X] complete graph methods
- [x] Update endpoint 
- [x] Ensure all test pass

### DIFF
--- a/ast/src/lang/graphs/neo4j_graph.rs
+++ b/ast/src/lang/graphs/neo4j_graph.rs
@@ -965,69 +965,13 @@ impl Neo4jGraph {
         txn_manager.execute().await
     }
     pub async fn create_filtered_graph_async(&self, _final_filter: &[String]) -> Result<Self> {
+        //Not needed for Neo4j
         Ok(self.clone())
     }
 
     pub async fn extend_graph_async(&mut self, other: Self) -> Result<()> {
-        let (other_nodes, other_edges) = other.get_graph_size_async().await?;
-        if other_nodes == 0 && other_edges == 0 {
-            debug!("Warning: Attempting to extend with an empty graph");
-            return Ok(());
-        }
-
-        let target_connection = self.ensure_connected().await?;
-        let source_connection = other.get_connection();
-        let mut txn_manager = TransactionManager::new(&target_connection);
-
-        let nodes_query = "MATCH (n) RETURN n, labels(n)[0] as node_type";
-        let mut result = source_connection.execute(query(nodes_query)).await?;
-        while let Some(row) = result.next().await? {
-            if let (Ok(node), Ok(node_type_str)) =
-                (row.get::<neo4rs::Node>("n"), row.get::<String>("node_type"))
-            {
-                if let Ok(node_type) = NodeType::from_str(&node_type_str) {
-                    let node_data = NodeData::try_from(&node).unwrap_or_default();
-                    txn_manager.add_node(&node_type, &node_data);
-                }
-            }
-        }
-
-        let edges_query = "MATCH (source)-[r]->(target)
-                           RETURN source, type(r) as edge_type, r, target,
-                                  labels(source)[0] as source_type,
-                                  labels(target)[0] as target_type";
-        let mut result = source_connection.execute(query(edges_query)).await?;
-        while let Some(row) = result.next().await? {
-            if let (
-                Ok(source_node),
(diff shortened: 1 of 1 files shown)


fayekelmith:
### Thought process for graph updates:

Code:
```rust
pub async fn update_incremental(
        &mut self,
        repo_url: &str,
        repo_path: &str,
        current_hash: &str,
        stored_hash: &str,
    ) -> Result<(u32, u32)> {
        let revs = vec![stored_hash.to_string(), current_hash.to_string()];
        if let Some(modified_files) = check_revs_files(repo_path, revs.clone()) {
            info!(
                "Processing {} changed files between commits",
                modified_files.len()
            );

            if !modified_files.is_empty() {
                let subgraph_repos = Repo::new_multi_detect(
                    repo_path,
                    Some(repo_url.to_string()),
                    modified_files.clone(),
                    Vec::new(),
                )
                .await?;

                let mut subgraph = Neo4jGraph::default();
                subgraph.connect().await?;
                subgraph.clear().await?;

                for repo in &subgraph_repos.0 {
                    let repo_subgraph = repo.build_graph_inner::<Neo4jGraph>().await?;
                    subgraph.extend_graph_async(repo_subgraph).await?;
                }

                self.graph.extend_graph_async(subgraph).await?;
                self.graph.create_indexes().await?;

                let (nodes_after, edges_after) = self.graph.get_graph_size();
                info!(
                    "Updated files: total {} nodes and {} edges",
                    nodes_after, edges_after
                );
            }
        }
```

- Get modified files
- Initialized a subgraph : this is the subgraph that will contain all the new nodes and edges across all the repositories (by language) of the files changed
- loop through the repositories (by Programming language) and build the graphs with Neo4jGraph and then extend the subgraph
- Now extend the subgraph to the main graph. The complete sync.

Evanfeenstra:
Im trying to understand this

build_graph should link all edges automatically right? Within the logic of ast

the purpose of the neo4j Graph implementation is to hook into the existing edge finding logic

fayekelmith:
> Im trying to understand this
> 
> build_graph should link all edges automatically right? Within the logic of ast
> 
> the purpose of the neo4j Graph implementation is to hook into the existing edge finding logic

I see, I failed to leverage that. 

I've simplified the implementation to use just `build_graph` to update the graph

ast/src/lang/graphs/graph_ops.rs
```diff
@@ -44,55 +44,29 @@ impl GraphOps {
...
-                    let (nodes_after, edges_after) = self.graph.get_graph_size();
-                    info!(
-                        "Updated files: added {} nodes and {} edges",
-                        nodes_after, edges_after
-                    );
+                for repo in &subgraph_repos.0 {
+                    self.graph = repo.build_graph_inner::<Neo4jGraph>().await?;
                 }
+
+                self.graph.create_indexes().await?;
```
Evanfeenstra, line 63: indexes dont need to be created each time. But we do need to update the `token_count` fields whenever new nodes are added

fayekelmith: sir, when creating the node, I added the token count directly... like this 

```rust
 let token_count = calculate_token_count(&self.node_data.body).unwrap_or(0);
 boltmap_insert_int(&mut properties, "token_count", token_count);
```

does this work?

Evanfeenstra:
> > Can you explain `create_filtered_graph_async`? It looks like its querying nodes from neo4j and then just putting them right back in neo4j? I dont understand
> 
> The idea is to copy the nodes and edges from the existing graph of the modified_files/filtered_files into a new Graph which is used by `filter_by_revs` in `builder.rs` to build the graph by processing only modified files.
> 
> I updated the code to ensure that the instance of the Database created for `filtered_graph` is always unique.
> 
> This logic is similar to what we have in `ArrayGraph` and `BTreeMapGraph`

Im not sure we need that at all?

The logic is totally different from ArrayGraph/BTreeMapGraph! Those ones have a 2-step process where we need to return the Graph from `build_graph_inner`, in order to upload to neo4j

Neo4JGraph does not have that 2-step process! Instead, we are updating the neo4j directly during build_graph_inner, correct? In your `update_incremental` function, you are not doing anything with the result of `build_graph_inner` right?

Evanfeenstra:
ok, for neo4j filter_by_revs can just return the full graph since its not needed

Issue #211, opened 2025-06-16
Ingestion for Sphinx Apps (ANDROID/KOTLIN)


tomsmith8:
## Source Code
​.kt​
​.kts​
​.java ​

## Gradle
​build.gradle​
​build.gradle.kts​
​settings.gradle​
​gradle.properties​
​.gradle​

## Manifest and Configuration
​AndroidManifest.xml​
​.xml​
​.properties​
.json, .yaml, .yml
proguard-rules.pro
keystore

## UI
​.xml​
​.strings​

## Dependencies Managemente
.iml
.idea (folder)
.gradlew, gradlew.bat
.lock, .versions.toml

## Third party
.json
​.db​, ​.sqlite​
.hilt
.graphql, .gql
.proto
.ico, .icns

tomastiminskas:
@Evanfeenstra updated info. In this case just considering the extensions should be enough. Maybe it's already done like this. We found missing files in the iOS repo, not in Android

## Source Code
Extension: ​`.kt​`, ​`.kts`​, `​.java ​`

## Gradle
Extension: `.gradle`, `.gradle.kts​`, `.properties​`

## Manifest and Configuration
Extension: `.xml`, ​`.properties​`, `.json`, `.yaml`, `.yml`, `.pro`

## UI
Extension:  ​`.xml​`

## Third party
Extensions: `.json`​, `.db`​, `​.sqlite​`, `.hilt`

PR #226, opened 2025-06-19
Adding files to Kotlin Implementation

fayekelmith:
Addressing #211


fayekelmith:
@Evanfeenstra 

What do you think about adding a new node_type `Config` to account for all the configuration files in big repos: 
Dockerfile, .conf, .yaml, .json, etc. 

They won't have direct edges with the rest of the repo except for Repo - CONTAINS -> Config  but I think it would be very useful context for LLMs if they are isolated from the sea of `File` nodes. 


Also, please is this PR headed in the right direction?

Issue #244, opened 2025-06-24
Function - Contains -> Var


fayekelmith:
The [Function] - Contains -> Var edge is not robust with LSP. 

Bench mark: `react.rs` test:

```rust
    let initial_state_var = variables
        .iter()
        .find(|v| {
            v.name == "initialState"
                && normalize_path(&v.file) == "src/testing/react/src/components/Person.tsx"
        })
        .map(|n| Node::new(NodeType::Var, n.clone()))
        .expect("initialState variable not found in Person.tsx");

    let use_store_fn = functions
        .iter()
        .find(|f| {
            f.name == "useStore"
                && normalize_path(&f.file) == "src/testing/react/src/components/Person.tsx"
        })
        .map(|n| Node::new(NodeType::Function, n.clone()))
        .expect("useStore hook not found in Person.tsx");

   assert!(
        graph.has_edge(&use_store_fn, &initial_state_var, EdgeType::Contains),
        "Expected useStore to contain initialState variable"
    );
```

WITH LSP turned on.

PR #246, opened 2025-06-25
fix collect var calls in function

fayekelmith:
address #244


ast/src/lang/parse.rs
```diff
@@ -1422,8 +1423,10 @@ impl Lang {
                 continue;
             }
 
-            if let Some(var) =
-                graph.find_node_by_name_and_file_end_with(NodeType::Var, target_name, &target_file)
```
Evanfeenstra, line 1426: why does `find_node_by_name_and_file_end_with` not work here?

PR #253, opened 2025-06-27
Angular nested pages

fayekelmith:
Created nested pages for Angular. 

Page[HTML] - Renders -> Page[HTML]


Evanfeenstra:
is this finding stuff like `<app-my-component>` in the HTML?

fayekelmith:
> is this finding stuff like `<app-my-component>` in the HTML?

Yes sir. I will test it with a real repo and share the results.

Issue #255, opened 2025-06-27
Next.js routes


Evanfeenstra:
Next.js routes are derived from the file path, not a string

The function names are like GET or POST, but the `Endpoint` node need be to captured from the file path

You'll have to parse the package.json to see if its using Next.js i guess...

PR #256, opened 2025-06-28
Support for nextjs

fayekelmith:
- [ ] Add comprehensive test file
- [ ] Setup specific testing 
- [ ] update queries till tests pass 

Addressing #255


fayekelmith:
@Evanfeenstra please can you review this PR?

ast/src/lang/queries/react.rs
```diff
@@ -442,8 +442,10 @@ impl Stack for ReactTs {
         }
     }
     fn update_request(&self, nd: &mut NodeData) {
-        if let Some(normalized_path) = normalize_frontend_path(&nd.name) {
-            nd.name = normalized_path;
+        if nd.name.starts_with('/') || nd.name.starts_with("api/") {
```
Evanfeenstra, line 445: can you explain this fix?

ast/src/builder/core.rs
```diff
@@ -621,6 +622,8 @@ impl Repo {
             info!("=> got {} function calls", i);
         }
 
+        linker::link_api_nodes(graph)?;
```
Evanfeenstra, line 625: `link_api_nodes` is also in Repos::build_graphs_inner... are we using it twice?

ast/src/lang/queries/react.rs
fayekelmith: It normalizes only the API endpoint names that are in the `/api` folder. I updated this because I noticed that if we have a purely React file with some Typescript in it, and then happen to have a request like `${api.host}/person}` it would normalize the `${api.host}` part of the name, which is incorrect/unnecessary. However, it ought to do so for all Nextjs files. And the routes are always in the `/api` directory.

ast/src/builder/core.rs
fayekelmith: the other `link_api_nodes` is in `Repos::build_graphs_inner` it only works when we are building graphs from Multiple repos. In this case we have just stepped up the capabilities of the React Queries to accomodate NextJs, so we have just one language. Since it's we have `Request` and `Endpoints` witihin thesame language, we had to call it in `Repo::build_graph_inner`; the other one is `Repos::build_graphs_inner`

fayekelmith: alright sir, I'll make the issue and make the update

Issue #258, opened 2025-06-30
services endpoint


Evanfeenstra:
GET /services
 - look through pkg_files to extract scripts (install, build, start, test), env vars

To generate a .devcontainer (pm2 and docker-compose files)

```json
[
  {
     name:"sphinx-tribes",
     language:"go",
     dev:true,
     scripts: {
       "install": ...,
       "build": "go build"
       "start": "./sphinx-tribes"
       "test" ...
     },
     env: [
       "DATABASE_URL", "JWT_KEY"
     ]
  },
  {
     name:"sphinx-tribes-frontend",
     language:"react",
     dev:true,
     scripts: {
       "install": "npm install",
       "start": "npm run dev"
       "test": "npm test"
     },
  },
  {
     name:"postgres",
  }
]
```

PR #269, opened 2025-07-01
/service endpoint

fayekelmith:
Addressing #258 

Building a /service endpoint in `mcp`


Evanfeenstra (commented):
What about making an `abstract Class ServiceParser` and then implement if with language-specific TsParser, GoParser, RubyParser, RustParser. With methods like:
- `env_names`: like `process.env`
- `env_regex`
- `pkg_file_name`
- `build`: (returning a Service type)

Evanfeenstra:
> 
> what do you think about it @Evanfeenstra ?

Lets keep it simple and keep it in the ts side. This will end up being messy and very flexible. That's why I dont want it at ingestion-time

PR #293, opened 2025-07-04
fix: wait for rust server

fayekelmith:
Fixing Wait rust server logic for github workflow


fayekelmith:
@Evanfeenstra this should solve the `Waiting for Rust Server` failure on the `e2e-standalone` GitHub workflow... please do you approve of it?

PR #311, opened 2025-07-07
fix: add explicit port to Neo4j connection and clean up docker-compose

thehimmat:
## Summary
- Add port to default hostname instead of hardcoding to prevent production breakage
- Remove obsolete version attribute from neo4j.yaml docker-compose file

## Changes
- Fixed Neo4j connection URI to use `localhost:7687` as default instead of appending `:7687`
- Handles both development (`localhost:7687`) and production (`neo4j.sphinx:7687`) environments correctly
- Removed obsolete `version: "3.8"` from docker-compose file

## Test plan
- [x] Verify Neo4j connection works in development without "no routing servers available" error
- [x] Confirm production won't break (NEO4J_HOST=neo4j.sphinx:7687 won't become neo4j.sphinx:7687:7687)
- [x] Verify docker-compose runs without version attribute warning
- [x] Confirm server starts successfully with Neo4j connection

### DIFF
--- a/mcp/src/graph/neo4j.ts
+++ b/mcp/src/graph/neo4j.ts
@@ -26,7 +26,7 @@ class Db {
   private driver: Driver;
 
   constructor() {
-    const uri = `neo4j://${process.env.NEO4J_HOST || "localhost"}:7687`;
+    const uri = `neo4j://${process.env.NEO4J_HOST || "localhost:7687"}`;
     const user = process.env.NEO4J_USER || "neo4j";
     const pswd = process.env.NEO4J_PASSWORD || "testtest";
     console.log("===> connecting to", uri, user, pswd);


mcp/src/graph/neo4j.ts
```diff
@@ -26,7 +26,7 @@ class Db {
   private driver: Driver;
 
   constructor() {
-    const uri = `neo4j://${process.env.NEO4J_HOST || "localhost"}`;
+    const uri = `neo4j://${process.env.NEO4J_HOST || "localhost"}:7687`;
```
Evanfeenstra, line 29: dont hardcode like this... on production we set NEO4J_HOST = `neo4.sphinx:7687`. So this would break 

You can add the port to the default if you want

const uri = `neo4j://${process.env.NEO4J_HOST || "localhost:7687"}`;

PR #328, opened 2025-07-08
Prevent self loops

fayekelmith:
Preventing self loops


ast/src/lang/call_finder.rs
```diff
@@ -24,24 +37,32 @@ pub fn func_target_file_finder<G: Graph>(
...
+    graph: &G,
+    source_start: usize,
+    current_file: &str,
+) -> Option<(String, usize)> {
     let mut target_files_starts = Vec::new();
     let nodes = graph.find_nodes_by_name(NodeType::Function, func_name);
     for node in nodes {
         // NOT empty functions (interfaces)
-        if !node.body.is_empty() {
+        if !node.body.is_empty() && current_file != node.file {
```
Evanfeenstra, line 63: node.body.is_empty() needs to be in both arms

PR #340, opened 2025-07-09
fix: data validation

### LINKED ISSUE #295: bug: Data Validation
Validate that the link added to the UI for ingesting data or via env is a Github link. Panic otherwise. 

This is both for the web UI and CLI

Image

fayekelmith:
Closes #295


standalone/static/app.js
```diff
@@ -62,22 +62,34 @@ const App = () => {
...
+      setRepoUrlError("");
+    }
   };
   const handleUsernameChange = (event) => {
     setUsername(event.target.value);
   };
   const handlePatChange = (event) => {
     setPat(event.target.value);
   };
+  function isValidGithubUrl(url) {
```
Evanfeenstra, line 84: this is too strict. what if the user copies from the url bar like `https://github.com/stakwork/stakgraph` ... it doesnt end with .git?

PR #345, opened 2025-07-10
Clean ups & fixes

fayekelmith:


Evanfeenstra:
we cant remove 

filter_by_revs

until we have everything working. we have multiple repos to test and migrate to the new /sync method. After its all working we can remove it

PR #353, opened 2025-07-10
multi language support

fayekelmith:
Permit multiple languages Neo4jGraph. draft1


fayekelmith:
@Evanfeenstra, this is what I was talking about before I lost connection.

Something like this.

PR #358, opened 2025-07-11
Improve Staktrak UI and test generation quality

jessejackson85:
## Changes

This PR enhances the Staktrak test recording tool with improved UI elements and better test output quality:

### UI Improvements
- Converted Start/Stop Recording into a single toggle button that changes text and color based on state
- Converted Assertion Mode/Interaction Mode into a toggle button with appropriate styling
- Added popup feedback messages during text selection and mode changes

### Test Generation Improvements
- Removed waitForTimeout calls before assertions in generated tests
- Added 500ms debounce for mouseUp events to prevent duplicate assertions from multi-clicks (like triple-clicks)
- Improved selector generation with cleaner output
- Removed staktrak-selection-active class from generated selectors
- Increased click filtering threshold to 1000ms to better handle triple-clicks
- Implemented better selector matching to filter out selection-related clicks

### Technical Details
- Modified index.html to handle button toggling and improved click filtering
- Updated staktrak.js with better debounce handling and selector generation
- Enhanced playwright-generator.js to produce cleaner test output without unnecessary waits

## Testing

The changes have been tested with various click patterns including:
- Single clicks for normal interactions
- Double and triple clicks for text selection
- Multiple assertions on the same element

The generated tests are now cleaner, more focused on meaningful interactions, and don't include unnecessary waitForTimeout calls before assertions.

## Demo:

https://www.loom.com/share/7259997912da4882ab59c927515b20e6

### DIFF
--- a/.github/workflows/publish-standalone.yml
+++ b/.github/workflows/publish-standalone.yml
@@ -32,6 +32,9 @@ jobs:
           file: Dockerfile.standalone
           platforms: linux/amd64,linux/arm64
           push: true
+          # caching
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
           tags: |
             ghcr.io/stakwork/stakgraph-standalone:${{ github.ref_name }}
             ghcr.io/stakwork/stakgraph-standalone:latest
--- a/Dockerfile.standalone
+++ b/Dockerfile.standalone
@@ -1,29 +1,43 @@
-
-# Use rust as the build environment
 FROM rust:1.87.0 AS builder
 
-# Create app directory
 WORKDIR /app
 
-# Copy the srcs and Cargo.tomls
-COPY ast/src ast/src
+# Create a temporary workspace with only the 3 crates we need
+RUN echo '[workspace]' > Cargo.toml && \
+    echo 'members = ["ast", "lsp", "standalone"]' >> Cargo.toml && \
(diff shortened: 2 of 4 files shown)


mcp/docs/staktrak/staktrak.js
```diff
@@ -306,23 +319,34 @@ var userBehaviour = (function () {
         /^(P|H1|H2|H3|H4|H5|H6|SPAN|DIV|LI|TD|TH|BUTTON|LABEL)$/i
       )
     ) {
-      return element.tagName.toLowerCase();
+      if (element.textContent && element.textContent.trim()) {
```
Evanfeenstra, line 322: what is this? why return the tagName if there is text?

Evanfeenstra (commented):
be careful with the getElementSelector function and make sure it works for all cases, think about very complex apps... 

could it be a performance problem to use querySelectorAll in getElementSelector?

mcp/docs/staktrak/staktrak.js
jessejackson85: @Evanfeenstra, The code was incorrectly assuming elements with text content could be uniquely identified by just their tag name (like 'p' or 'div'), which isn't true in most cases. I removed this flawed logic to create more reliable selectors.

jessejackson85:
@Evanfeenstra Fixed

Issue #376, opened 2025-07-14
ingest other repo


Evanfeenstra:
the standalone /ingest endpoint should accept a new repo, and only delete changed repos

So like first you can ingest `sphinx-tribes`... and then later ingest `sphinx-tribes-frontend`, and it wont delete any nodes from `sphinx-tribes`

Also, does the `commit` param accept comma-separated list of commits one for each repo?

PR #377, opened 2025-07-14
fix: Ingest multiple

fayekelmith:
First part of solution to #376


standalone/src/handlers.rs
```diff
@@ -229,9 +229,6 @@ pub async fn ingest(
     let mut graph_ops = GraphOps::new();
     graph_ops.connect().await?;
 
-    info!("Clearing old data...");
-    graph_ops.graph.clear().await?;
```
Evanfeenstra, line 233: you still need to clear... but only the one you ingested

so use file STARTS WITH

Issue #136, opened 2025-05-20
ruby partials


Evanfeenstra:
we need RENDERS edges from .haml, .rb, .rb etc for "partials", which are like partial HTML sections

the erb treesitter is not very good... i think we need to search strings for `partial: `, or use LSP

jordan-ae:
Hello can I work on this?

PR #378, opened 2025-07-14
Improve Ruby partial rendering detection using regex patterns

jessejackson85:
**PR Description:**

This PR improves the detection of Ruby on Rails partials by using regex patterns instead of tree-sitter. It extracts partial names from template files (.erb, .haml, .slim) and Ruby controller files (.rb) to create accurate RENDERS edges in the dependency graph.

Key changes:
- Add regex-based partial extraction for more reliable detection
- Support standard Rails partial conventions (underscore prefix, shared directories)
- Handle various render syntaxes (`render partial:`, `<%= render`, etc.)
- Expand supported template formats
- Improve path resolution to find partials in different locations

Fixes the issue where partial relationships weren't properly detected in Rails applications.

**Closed**  #136

### DIFF
--- a/ast/src/lang/queries/ruby.rs
+++ b/ast/src/lang/queries/ruby.rs
@@ -10,6 +10,7 @@ use std::collections::BTreeMap;
 use std::path::Path;
 use tracing::debug;
 use tree_sitter::{Language, Parser, Query, Tree};
+use regex::Regex;
 
 pub struct Ruby(Language);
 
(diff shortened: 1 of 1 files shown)


jessejackson85:
@Evanfeenstra Please review this PR.

jessejackson85:
@Evanfeenstra I've pushed the newly added feature here. Please review it.

PR #379, opened 2025-07-14
fix: add repo with specific revs

fayekelmith:
Addressing #376


Evanfeenstra:
once we remove `filter_by_revs` we can remove `revs` on Repo. Then remove revs completely

PR #391, opened 2025-07-15
Fix/lsp cleanup

### LINKED ISSUE #390: post clone cmd
post clone cmd (npm install) should ONLY run if the package.json changed

so /sync will be faster

fayekelmith:
Closes #390


fayekelmith:

PR #402, opened 2025-07-16
fixed demorepo test

fayekelmith:


fayekelmith:
@Evanfeenstra , this is the demorepo test update

PR #418, opened 2025-07-18
test: local e2e test

fayekelmith:
Setting up a local e2e test. An integrated test.


fayekelmith:
The tests are not entirely reliable yet because, although I sort them, it's hard to maintain structural integrity (for the edges) and consistently pass. Sometimes the context is correct, but the tests fail. It's because sorting is not yet efficient. 

@Evanfeenstra, please do you have any suggestions on how to solve this without losing the structure?

standalone/tests/maps/actual-map-response-two.html
```diff
@@ -0,0 +1,15 @@
+<pre>
+Function: NewPerson (22)
+├── Datamodel: Person (30)
+├── Class: database (9)
+├── Function: CreatePerson (124)
+├── Endpoint: POST /person (8)
```
Evanfeenstra, line 6: this looks like NewPerson has an edge with `Endpoint: POST /person`??

Shouldn't it be like `NewPerson` -> `Request` -> `Endpoint`?

or is this html sorted somehow ?

fayekelmith: After looking at it again, something more fundamentally wrong is actually the issue. 

I queried `NewPerson` which is a function with the same name existing for both programming languages. From my observation of the results, it mixed both results. e.g

We have both a relationship with `Class: database` , `Endpoint: POST /person`,  `Page: /new-person`  and `Request: POST ${api.host}/person` 

I think the results returned are a combination of both maps.

Evanfeenstra:
> The tests are not entirely reliable yet because, although I sort them, it's hard to maintain structural integrity (for the edges) and consistently pass. Sometimes the context is correct, but the tests fail. It's because sorting is not yet efficient.
> 
> @Evanfeenstra, please do you have any suggestions on how to solve this without losing the structure?

We need to sort it before its returned from the server (before its formatted with the "archy" lib). Like a heirarchical sort:
```
Function: App (85)
├─┬ Function: People (155)
│     ├── Request: GET ${api.host}/people (7)
|      ├── Datamodel: Person (19)
└┬ Function: NewPerson (315)
    ├── Function: FormContainer (60)
    ├── Function: Apple (60)
  ```

  should be sorted as
  
  ```
  Function: App (85)
└┬ Function: NewPerson (315)
    ├── Function: Apple (60)
    ├── Function: FormContainer (60)
├─┬ Function: People (155)
|      ├── Datamodel: Person (19)
│     ├── Request: GET ${api.host}/people (7)

  ```
  
  Because `Function: NewPerson` comes before `Function: People` alphabetically
  and `Function: Apple` comes before its sibling `Function: FormContainer`, etc
  
 only "siblings" should be sorted, recursively down the whole tree
  
  Since every edge has the same "weight" in the graph, there is no specific meaning to the order of the graph right now. So we can sort it at every level

fayekelmith:
> > The tests are not entirely reliable yet because, although I sort them, it's hard to maintain structural integrity (for the edges) and consistently pass. Sometimes the context is correct, but the tests fail. It's because sorting is not yet efficient.
> > @Evanfeenstra, please do you have any suggestions on how to solve this without losing the structure?
> 
> We need to sort it before its returned from the server (before its formatted with the "archy" lib). Like a heirarchical sort:
> 
> ```
> Function: App (85)
> ├─┬ Function: People (155)
> │     ├── Request: GET ${api.host}/people (7)
> |      ├── Datamodel: Person (19)
> └┬ Function: NewPerson (315)
>     ├── Function: FormContainer (60)
>     ├── Function: Apple (60)
> ```
> 
> should be sorted as
> 
> ```
> Function: App (85)
> └┬ Function: NewPerson (315)
>   ├── Function: Apple (60)
>   ├── Function: FormContainer (60)
> ├─┬ Function: People (155)
> |      ├── Datamodel: Person (19)
> │     ├── Request: GET ${api.host}/people (7)
> ```
> 
> Because `Function: NewPerson` comes before `Function: People` alphabetically and `Function: Apple` comes before its sibling `Function: FormContainer`, etc
> 
> only "siblings" should be sorted, recursively down the whole tree
> 
> Since every edge has the same "weight" in the graph, there is no specific meaning to the order of the graph right now. So we can sort it at every level

Okay, that makes sense, sir. 
#422 solves this issue.

Evanfeenstra:
ok i merged 422, rebase on main and see if it helps

PR #430, opened 2025-07-21
fix: react return types

fayekelmith:
- Adding the Return type edge


Evanfeenstra (commented):
so this didn't change any test? Why aren't there more edges collected?

PR #431, opened 2025-07-21
feat: progress status api

fayekelmith:
- Establish state management across crates. 
-


Evanfeenstra:
I think it would be much simpler to pass an `Option<Sender<T>>` (where Sender is tokio::sync::mpsc::Sender) from standalone into ast build_graph_inner

PR #436, opened 2025-07-22
refactor: core.rs

fayekelmith:
- Attempt to optimize the graph building process by reducing the number of times files are parsed. 

## Goal
- Combine all nodes that are independent under one files traversal

### DIFF
--- a/ast/src/builder/core.rs
+++ b/ast/src/builder/core.rs
@@ -46,9 +46,7 @@ impl Repo {
         self.process_libraries(&mut graph, &filez)?;
         self.process_import_sections(&mut graph, &filez)?;
         self.process_independent_nodes(&mut graph, &filez)?;
-        // self.process_variables(&mut graph, &filez)?;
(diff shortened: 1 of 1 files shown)


ast/src/builder/core.rs
```diff
@@ -45,9 +45,8 @@ impl Repo {
 
         self.process_libraries(&mut graph, &filez)?;
         self.process_import_sections(&mut graph, &filez)?;
-        self.process_variables(&mut graph, &filez)?;
-        self.process_classes(&mut graph, &filez)?;
-        self.process_instances_and_traits(&mut graph, &filez)?;
+        self.process_independent_nodes(&mut graph, &filez)?;
```
Evanfeenstra, line 48: im not sure i understand what this changes. ideally we have each node type in its own function, so its easier to track progress and read

fayekelmith:
To be attempted in a different PR.

PR #444, opened 2025-07-23
feat: Class - IMPLEMENTS -> Trait

fayekelmith:
- Add the implements edge
- Implement it on Rust

### DIFF
--- a/ast/src/lang/parse/format.rs
+++ b/ast/src/lang/parse/format.rs
@@ -253,6 +253,26 @@ impl Lang {
         })?;
         Ok(inst)
     }
+
+    pub fn format_implements(
+        &self,
+        m: &QueryMatch,
+        code: &str,
+        q: &Query,
+    ) -> Result<(String, String)> {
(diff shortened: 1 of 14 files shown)


ast/src/lang/parse/format.rs
```diff
@@ -54,7 +55,28 @@ impl Lang {
...
             Ok(())
         })?;
-        Ok((cls, associations))
+
+        if self.lang.filter_by_implements() {
+            if let Some(implements_query) = self.lang.implements_query() {
+                let implements_q = self.lang.q(&implements_query, &NodeType::Class);
+                let tree = self.lang.parse(code, &NodeType::Class)?;
+                let mut cursoe = QueryCursor::new();
+                let mut matches = cursoe.matches(&implements_q, tree.root_node(), code.as_bytes());
```
Evanfeenstra, line 64: i wonder if we can directly query the Class node here, instead of the entire file

`node` instead of `tree.root_node()`

it should be much more efficient

Issue #452, opened 2025-07-25
rust enums


Evanfeenstra:
are rust enums considered DataModel?

They should be

Also typescript enum

PR #455, opened 2025-07-25
fix: Rust enums as Data model

fayekelmith:
Addressing #452 

added rust ENUMS as DataModels and Class (if they have an impl) it's based on #444

### DIFF
--- a/ast/src/lang/parse/format.rs
+++ b/ast/src/lang/parse/format.rs
@@ -253,6 +253,26 @@ impl Lang {
         })?;
         Ok(inst)
     }
+
+    pub fn format_implements(
+        &self,
+        m: &QueryMatch,
+        code: &str,
+        q: &Query,
+    ) -> Result<(String, String)> {
+        let mut class_name = String::new();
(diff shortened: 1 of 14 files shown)


ast/src/lang/parse/format.rs
```diff
@@ -54,7 +55,28 @@ impl Lang {
...
             Ok(())
         })?;
-        Ok((cls, associations))
+
+        if self.lang.filter_by_implements() {
+            if let Some(implements_query) = self.lang.implements_query() {
+                let implements_q = self.lang.q(&implements_query, &NodeType::Class);
+                let tree = self.lang.parse(code, &NodeType::Class)?;
+                let mut cursoe = QueryCursor::new();
+                let mut matches = cursoe.matches(&implements_q, tree.root_node(), code.as_bytes());
```
Evanfeenstra, line 64: cant we use the `node` already collected from the CLASS? Instead of `tree.root_node()`

This way, its running queries over the entire file... but we only need to query the class text right?

PR #456, opened 2025-07-25
fix: enum and types as DataModels for TS based languags

fayekelmith:
Address #455 for Typscritpt, Rust, and Angular. 

updated their queries to capture Enum and Types as DataModels as well.

### DIFF
--- a/ast/src/lang/queries/angular.rs
+++ b/ast/src/lang/queries/angular.rs
@@ -195,10 +195,17 @@ impl Stack for Angular {
     fn data_model_query(&self) -> Option<String> {
         Some(format!(
             r#"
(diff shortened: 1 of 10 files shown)


fayekelmith:
> is there more testing we can add here?
> 
> every possible combination should be tested...
> 
> * interface with no methods
> * interface with methods
> * type
> * class `implements` interface
> * etc etc etc, there are probably a lot right?

I'll update the tests for identifying different DataModels.

For class - Implements -> Interface tests, I'll add them in a different PR. This PR is mostly about just Data Model Queries.

Evanfeenstra (approved):

PR #457, opened 2025-07-25
feat: go implements edge

fayekelmith:
Implement the `Class` (Struct) - IMPLEMENTS -> `Trait` (Interface) Edge in Go.


fayekelmith:
Go is a special case because of how `Structs` implement `Interfaces`, need to rethink the strategy here.

Issue #458, opened 2025-07-26
TS: DataModel vs Trait Query


fayekelmith:
In TS: 
One form of a DataModel is : An interface without methods

An interface/type with methods is a Trait. 

To capture a Trait we use the Treesitter query:

```rust
    fn trait_query(&self) -> Option<String> {
        Some(format!(
            r#"
            [
                (interface_declaration
                    name: (type_identifier) @{TRAIT_NAME}
                    body: (interface_body
                        (method_signature)+
                    )
                )
                (type_alias_declaration
                    name: (type_identifier)@{TRAIT_NAME}
                    value: (object_type
                            (method_signature)+
                        )
                )
            ]@{TRAIT}
            "#
        ))
    }
```
It's easy to ensure that an interface MUST have a method in it. `(method_signature)+`

But it's much harder to skip an interface if it contains a method: 

I've tried:

```sql
 (interface_declaration
        name: (type_identifier) @struct-name
        body: (interface_body

            (method_signature) @method
            (#is_not? @method _)

        )
    ) @struct
```

CLAUDE suggested:

```
(interface_declaration
  name: (type_identifier) @struct-name
  body: (interface_body
    !(_ (method_signature))
  )
) @struct
``` 
But it doesn't work. It creates an error.

PR #459, opened 2025-07-26
feat: implements edge for Typescript 

fayekelmith:
This PR:
- Adds the implements edge from Typescript classes to Traits (Interfaces with methods). 
- updates test to test the solutions 

Also, updates tests to serve standard for the solution of #458 

We should have:
8 DataModels ideally and 2 Traits instead of 
10 DataModels and 2 Traits as we do now.


fayekelmith:
Based on the work from #444 and #456

Evanfeenstra (approved):

Issue #443, opened 2025-07-23
fix: Error propagation in standalone


fayekelmith:
If the repo is not cloneable or if any error occurs in the standalone server, it's not being handled in the frontend. e.g when Cloning a repo, if the repository doesn't exist, the UI is never informed of this error.

PR #463, opened 2025-07-27
fix: propagate errors to standalone

fayekelmith:
- Propagate errors correctly from the source
- Use the Errors Propagated to Update the UI
- Using the API endpoint to get errors

Addressing #443

### DIFF
--- a/ast/src/testing/nextjs/mod.rs
+++ b/ast/src/testing/nextjs/mod.rs
@@ -441,7 +441,7 @@ pub async fn test_nextjs_generic<G: Graph>() -> Result<()> {
 }
 
 #[cfg(all(feature = "neo4j", feature = "fulltest"))]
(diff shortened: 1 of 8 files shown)


fayekelmith:
Needed to update `new_clone_multi_detect` and consequently `clone_multi_detect` to accommodate `status_tx`, which I used to send errors to the UI to the frontend using the broadcast channel (as they occur). 

In the repo cloning and walking process, I updated the involved methods to report errors correctly; that is, the `run` command and the `git_clone` command. I intend to send these errors directly to the UI, where they will provide context of what went wrong.

fayekelmith:
Screenshot 2025-07-28 at 12 24 58 AM

For some reason, the error propagated from the Rust axum server does not display on the UIl even when it logs on the status

fayekelmith:
Screenshot 2025-07-30 at 2 35 10 PM

@Evanfeenstra, please review this PR

PR #470, opened 2025-07-28
Feat: nextjs pages

fayekelmith:
Addressing #467 

The Goal is:

- Create Nextjs Pages as: 
     - name: folder name
     - file: full path
     - body: relative path without `page.tsx`
Create Page - RENDERS ->  Functions (Components in the file). 

---
- Updated the implementation in Angular so that PAGE node and Render EDGE  addition accommodate a more language-agnostic implementation. 
- Update the tests for Angular. 
- Rebased to updated nextjs testing.

### DIFF
--- a/ast/src/builder/core.rs
+++ b/ast/src/builder/core.rs
@@ -618,25 +618,30 @@ impl Repo {
             page_count += extra_page_count;
 
             for pagepath in extra_pages {
-                if let Some(pagename) = get_page_name(&pagepath) {
+                if let Some((page_node, edges)) = self.lang.lang().extra_page_finder(
+                    &pagepath,
+                    &|name, filename| {
+                        graph.find_node_by_name_and_file_end_with(
+                            NodeType::Function,
+                            name,
+                            filename,
+                        )
+                    },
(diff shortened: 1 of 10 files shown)


Evanfeenstra:
other possibilities:

```tsx
const MyComponent = () => {
  return <div>Hello</div>
}

export default MyComponent
```

```tsx
export default () => {
  return <div>Hello</div>
}
```

```tsx
const MyComponent = () => {
  return <div>Hello</div>
}

export { MyComponent as default }
```

```tsx
import { NextPage } from 'next'

const MyComponent: NextPage = () => {
  return <div>Hello</div>
}

export default MyComponent
```

```tsx
import React from 'react'

const MyComponent: React.FC = () => {
  return <div>Hello</div>
}

export default MyComponent
```

```tsx
export default async function MyComponent() {
  const data = await fetch('...')
  return <div>Hello</div>
}
```

```tsx
interface Props {
  params: { id: string }
  searchParams: { [key: string]: string | string[] | undefined }
}

export default function MyComponent({ params, searchParams }: Props) {
  return <div>Hello</div>
}
```

PR #471, opened 2025-07-29
feat: add envs to /service reponse

fayekelmith:
The goal is to use a cloned repo, walk through it and use regex expressions to collect all the ENV files used in the code base.


fayekelmith:
Screenshot 2025-07-30 at 12 35 04 AM
Tested and works.

PR #492, opened 2025-08-03
fix: add cpp libraries

fayekelmith:
- [X] Updated the implementation to use `bash` and collect `CPP` libraries


fayekelmith:

sempervictus:
@fayekelmith - looks like this will at least partially catch CUDA code but since its ~CPP is there anything else needed to get coverage on that?

PR #496, opened 2025-08-04
test: leaderboard integrated test

fayekelmith:
Added the LeaderboardPage e2e integrated test script


fayekelmith:
> The Request nodes are not linked to the Endpoint nodes
> 
> Don't ingest hive
> 
> Ingest sphinx-tribes and sphinx-tribes-frontend at the same time (together in one request), so that the api nodes are linked

I think we'll have to use `/sync` rather than ingest. Ingest will build the graph as a `BTreemapGraph` and then upload; there won't really be a link between both repositories, because they were built and uploaded separately.

With ingest, we'll build the Graph as a `Neo4jGraph`, then we can link them. 


Either we sync, or maybe we can revisit `/ingest`s flow? 

@Evanfeenstra

PR #537, opened 2025-08-20
Get a kotlin LSP to run in the lsp crate

azzizt:
1. Kotlin LSP Integration & Docker Kotlin LSP is now running in the lsp crate: The identifier query was updated, and the LSP is being used for Kotlin analysis. Dockerfile: The base image includes the kotlin-language-server, so LSP-based analysis is available in the container.
2. Test Results: LSP vs Non-LSP Non-LSP (static analysis): Previously, the test expected 6 import edges and 175 contains edges. LSP-enabled: After enabling LSP and updating the identifier query, the test now finds 178 contains edges.

### DIFF
--- a/ast/src/lang/queries/kotlin.rs
+++ b/ast/src/lang/queries/kotlin.rs
@@ -12,6 +12,10 @@ impl Kotlin {
 }
 
 impl Stack for Kotlin {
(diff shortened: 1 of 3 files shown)


azzizt:
here is reopen  #542 #542

Issue #536, opened 2025-08-20
after every step of ast/builder/core.rs upload to neo4j


Evanfeenstra:
to have incremental sync

still using BTreeMap

```

        self.process_libraries(&mut graph, &filez)?;
        // update load to neo4j
        self.process_import_sections(&mut graph, &filez)?;
        // upload to neo4j 
        self.process_variables(&mut graph, &filez)?;
         // upload to neo4j 
        self.process_classes(&mut graph, &filez)?;
         // upload to neo4j 
        self.process_instances_and_traits(&mut graph, &filez)?;
         // upload to neo4j 
        self.process_data_models(&mut graph, &filez)?;
         // upload to neo4j 
        self.process_functions_and_tests(&mut graph, &filez).await?;
         // upload to neo4j 
        self.process_pages_and_templates(&mut graph, &filez)?;
        self.process_endpoints(&mut graph, &filez)?;
        self.finalize_graph(&mut graph, &filez, &mut stats).await?;
        let graph = filter_by_revs(
            &self.root.to_str().unwrap(),
            self.revs.clone(),
            graph,
            self.lang.kind.clone(),
// etc
```

PR #538, opened 2025-08-20
feature: incremental builds

fayekelmith:
Closes : #536


ast/src/builder/core.rs
```diff
@@ -30,31 +35,378 @@ impl Repo {
     }
     pub async fn build_graph_inner<G: Graph>(&self) -> Result<G> {
         let graph_root = strip_tmp(&self.root).display().to_string();
-        let mut graph = G::new(graph_root, self.lang.kind.clone());
+        let mut graph = G::new(graph_root.clone(), self.lang.kind.clone());
         let mut stats = std::collections::HashMap::new();
 
+        #[cfg(feature = "neo4j")]
+        let mut btree_graph = BTreeMapGraph::new(graph_root, self.lang.kind.clone());
```
Evanfeenstra, line 42: Why make a new Graph? Just use the graph

fayekelmith: the graph is generic, sir...and I need a concrete implementation of BtreeMap to upload to Neo4j

Evanfeenstra: why do you need concrete implementation? For example here is 
```
pub async fn upload_nodes_by_type(
        &mut self,
        btree_graph: &BTreeMapGraph,
        node_type: NodeType,
    ) -> Result<()> {
        let nodes = btree_graph.find_nodes_by_type(node_type.clone());
        if nodes.is_empty() {
            return Ok(());
        }

        let queries: Vec<(String, BoltMap)> = nodes
            .iter()
            .map(|node| add_node_query(&node_type, node))
            .collect();

        self.graph_ops.graph.execute_batch(queries).await
    }
    ```
    
    why cant that just take a `graph: &G`? find_nodes_by_type is available on all graphs
    
    Of course, for neo4jgraph we dont want to do this incremental stuff at all. So we can have a `do_incremental` method on Graph that just returns a boolean, so only for BTreeMap it is actually performed

PR #542, opened 2025-08-22
Get a kotlin LSP to run in the lsp crate

azzizt:
Kotlin LSP Integration & Docker Kotlin LSP is now running in the lsp crate: The identifier query was updated, and the LSP is being used for Kotlin analysis. Dockerfile: The base image includes the kotlin-language-server, so LSP-based analysis is available in the container.
Test Results: LSP vs Non-LSP Non-LSP (static analysis): Previously, the test expected 6 import edges and 175 contains edges. LSP-enabled: After enabling LSP and updating the identifier query, the test now finds 16 import edges 178 contains edges.


azzizt:
added the JDK and set JAVA_HOME in  Dockerfile is required by  the Kotlin Language Server. 
updated the tests eExpected edges . . @Evanfeenstra

azzizt:
updated dockerfile here sir #544

PR #570, opened 2025-08-30
feat: simpler implementation of batch upload

fayekelmith:
- loading


Evanfeenstra:
might be better to add a 
```rs
fn iter_nodes(&self) -> Box<dyn Iterator<Item = (&str, &Node)> + '_>;
```

than use downcast_ref like this? what do you think @fayekelmith ?

Issue #566, opened 2025-08-29
Add property date_added_to_graph in unix timestamp


tomsmith8:
## Task

Any time we add or sync a node in the graph, store date_added_to_graph in unix timestamp

example: `"date_added_to_graph": "1731665418.2726846"`

PR #572, opened 2025-09-01
add last modified to node

### LINKED ISSUE #567: Extend /graph endpoint
Currently we can get only one node type at a time. That means there is no way to get other node types, and connections between them. What we need to do is to return latest nodes of different types, and connections between them. 

possible solutions: 
- add latest param (based on https://github.com/stakwork/stakgraph/issues/566)
- add limit
- node_type should accept several types or none(return all)

fayekelmith:
Extending /graph endpoint 

Closes #567 #566

### DIFF
--- a/ast/src/lang/graphs/neo4j_utils.rs
+++ b/ast/src/lang/graphs/neo4j_utils.rs
@@ -78,7 +78,7 @@ impl NodeQueryBuilder {
     let query = format!(
         "MERGE (node:{}:{} {{node_key: $node_key}})
          ON CREATE SET node += $properties, node.date_added_to_graph = $now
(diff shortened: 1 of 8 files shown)


fayekelmith:

ast/src/lang/graphs/neo4j_utils.rs
```diff
@@ -75,14 +75,14 @@ impl NodeQueryBuilder {
...
-            ON CREATE SET node += $properties
-            ON MATCH SET node += $properties
-            Return node",
-            self.node_type.to_string(),
-            DATA_BANK,
-        );
+    let query = format!(
+        "MERGE (node:{}:{} {{node_key: $node_key}})
+         ON CREATE SET node += $properties, node.date_added_to_graph = $now
+         ON MATCH SET node.date_added_to_graph = CASE WHEN node.hash IS NULL OR node.hash <> $properties.hash THEN $now ELSE node.date_added_to_graph END, node += $properties
```
Evanfeenstra, line 81: why are you checking node.hash here?

PR #581, opened 2025-09-02
Feat: function capture v2

fayekelmith:
attempt to fix loss calls in #574


ast/src/lang/queries/react.rs
```diff
@@ -327,10 +313,11 @@ impl Stack for ReactTs {
                         )
                     )
                 )
-            )
-        ] @{FUNCTION_DEFINITION}"#
-        )
-    }
+            ) @{FUNCTION_DEFINITION}
```
Evanfeenstra, line 316: lets revert this to the original form

PR #595, opened 2025-09-05
feat: understanding endpoint and hint node type

### LINKED ISSUE #585: Project "Hints"
Phase 1

1. Pass in a question about the code base
2. Agentically explore the code to answer the question (use /explore route)
3. Save the output as a text chunk (we are calling these Hints)
4. You now have a question paired with a text summary
5. Store the question and answer as a node in the graph
6. Generate edges between this Q&A node to each file or function mentioned in the output
7. Vectorize just the question part the of text

When a user asks a question in Hive, we search the vectorized questions and pass back the plain text answer.

Phase 2
Ask the LLM to generate the 100 top questions to ask about a repo.

Phase 3
Expose the use of this to /explore
Explose /explore as an MCP for Goose/Agent

Phase 4
Cross repo version of this process

fayekelmith:
- [X] Add Hint NodeType
- [X] Create `GET /understanding`
- [x] Edge creation from Hint to Reference Nodes
- [x] Update /understanding to be more flexible

Closes #585

### DIFF
--- a/mcp/src/graph/neo4j.ts
+++ b/mcp/src/graph/neo4j.ts
@@ -1,4 +1,3 @@
-import { anthropic } from "@ai-sdk/anthropic";
 import neo4j, { Driver, Session } from "neo4j-driver";
 import fs from "fs";
 import readline from "readline";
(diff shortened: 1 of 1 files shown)


mcp/src/graph/neo4j.ts
```diff
@@ -512,6 +530,143 @@ class Db {
     }
   }
 
+  async create_hint(question: string, answer: string, embeddings: number[]) {
+    const session = this.driver.session();
+    const name = question.slice(0, 80);
+    const node_key = `Hint|:|${name}|:|hint://generated|:|0`;
```
Evanfeenstra, line 536: can you clean up the node key? i think we already have function to clean string (no special characters, strings, etc)

PR #599, opened 2025-09-06
feat:  interface components

### LINKED ISSUE #552: function interface
It would be useful to have just the function interface stored separately

Like just including the beginning text of a function (export, name, generics, arguments, return type)

Can we simply count the last character of the treesitter RETURN_TYPE, and take those first x character, and put them on a new meta field called "interface"

fayekelmith:
Closes #552


fayekelmith:
> are you sure this is the best way?
> 
> Each capture has a line/column right?
> 
> so just capturing the return type and taking the entire body up to the last character of the return type should be correct right? or if there is no return type, then up to the last argument...
> 
> I'm not sure, but that seems little simpler?

Indeed is a cleaner and simpler approach:

PR #603, opened 2025-09-06
fix: use ends with to match name by ending

fayekelmith:
- [X] Ensure that files are mapped to Hint. 

Screenshot 2025-09-07 at 1 37 35 AM

Test on demo repo: `curl -G 'http://localhost:3000/understanding' --data-urlencode 'question=Detailly, what does this repo achive?' --data-urlencode 'provider=google'`

### DIFF
--- a/mcp/src/graph/neo4j.ts
+++ b/mcp/src/graph/neo4j.ts
@@ -582,17 +582,29 @@ class Db {
     const schema = z.object({
       function_names: z
         .array(z.string())
-        .describe("functions or react components"),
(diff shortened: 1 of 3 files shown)


Evanfeenstra:
Please ingest hive and try this again

files like src/dir/file.ts will not match right? File needs to be a special search i think (matching file.file not file.name)

fayekelmith:
> Please ingest hive and try this again
> 
> files like src/dir/file.ts will not match right? File needs to be a special search i think (matching file.file not file.name)

Screenshot 2025-09-07 at 2 59 36 AM

PR #646, opened 2025-09-24
Refactor StakTrak data management architecture - move state from iframe to parent

thehimmat:
StakTrak Data Management Refactoring Summary

Original Goal: Move recording logic from StakTrak iframe to host environment for better architecture and reusable TypeScript library.

PR #646 Implementation: Successfully moved state management to host via RecordingManager class and event-based messaging. However, this created architectural debt with RecordingManager duplicated in both iframe and host bundles, plus extensive fallback logic for backward compatibility.

This Refactoring: Cleaned up the architecture by removing 305 lines of redundant code with zero functionality loss. Eliminated RecordingManager duplication, removed all fallback/compatibility code paths, and simplified to single message protocol.

Result: Clean architectural separation achieved - StakTrak iframe (144KB) handles pure event capture, playwright-generator host (14KB) contains RecordingManager and test generation logic. Single source of truth for recording state, improved maintainability, and foundation ready for future enhancements.

Technical Impact: -305 lines total, same user functionality, cleaner codebase, better bundle separation, and RecordingManager now ready for potential standalone packaging.

### DIFF
--- a/mcp/tests/app.js
+++ b/mcp/tests/app.js
@@ -336,11 +336,7 @@ const Staktrak = () => {
                   }
                 }
 
-                // Extract tag name (fallback)
-                const tagMatch = selector.match(/^([a-zA-Z]+)/);
-                if (tagMatch) return tagMatch[1];
-
-                // If selector is too complex, truncate it
+                // Use selector as fallback (truncate if too long)
                 if (selector.length > 25) {
                   return selector.substring(0, 22) + '...';
                 }


Evanfeenstra:
looking at the loom... 

- why is the complex replace logic inside the test text itself? can't you just replace first and have a clean url in the test?

- it looks like mouseup is also recorded as a "click" when you assert? that should be cleaned up

- the steps could show like "click id: my-id" (show the actual id/class/text value maybe)

- did u test this on hive?

thehimmat:
Thanks for reviewing! Addressed several of these issues in latest commits:

**URL normalization**: ✅ Fixed - URLs are now pre-normalized during action creation and stored as `normalizedUrl`, so the generated tests have clean URLs without complex replace logic.

**mouseup → click recording**: ✅ Should be resolved - the code has logic to skip click recording when in selection mode (`if (this.memory.selectionMode) return`). The mouseup listener is only for text selection/assertions, not click recording.

**Better step descriptions**: ✅ Improved - Latest commit enhances element descriptions to show actual ID/class/attribute values (e.g. "Click #my-id", "Type in [name=email]") with simplified fallback logic.

**Hive testing**: 🔄 Still need to test this on Hive specifically. Will do that next.

The recent commits focused on cleaning up the action description system and removing some brittle regex logic in favor of more reliable approaches.

thehimmat:
https://www.loom.com/share/4482b545ad64435094aeadc2058c6a36?sid=731df59d-f2c7-4852-8e1e-814df1d5164e

PR #650, opened 2025-09-25
Add file-based search to map endpoint as alternative to name search

aliraza556:
## How It Works
The endpoint now accepts three different ways to find your starting node:

**Option 1: By name (existing)**
```bash
GET /map?node_type=Function&name=authenticateUser
```

**Option 2: By file path (new!)**  
```bash
GET /map?node_type=Function&file=auth/login.ts
GET /map?node_type=Component&file=components/user
```

**Option 3: By ref_id (existing)**
```bash
GET /map?ref_id=abc-123-def
```

## Technical Details
- Uses `CONTAINS` matching for file paths (so `file=auth` finds `src/auth/login.ts`)
- Maintains backward compatibility - all existing queries work unchanged
- Added proper validation to ensure you provide either name+type, file+type, or ref_id
- Updated the Cypher query to handle the new file-based node lookup

### DIFF
--- a/mcp/src/graph/graph.ts
+++ b/mcp/src/graph/graph.ts
@@ -187,6 +187,7 @@ function formatEdge(edge: Neo4jEdge): string {
 export interface MapParams {
   node_type: string;
   name: string;
+  file: string;
   ref_id: string;
   tests: boolean;
   depth: number;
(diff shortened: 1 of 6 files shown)


Evanfeenstra:
@aliraza556 can you rebase on main, i changed the query: https://github.com/stakwork/stakgraph/commit/1cfc445f6d3cf225c15ace5c910a22e7bc66b817

PR #658, opened 2025-09-29
feat: single file bin setup

fayekelmith:
- [X] Single flle bin established
- [x] finish parsing logic
- [x] Create a print out logic

`cargo run --bin single -- path/to/file.ts`


Evanfeenstra (commented):
dont use `default_do_lsp` for this

Evanfeenstra:
Why did you delete default_do_lsp? Keep it for regular operations. But for this single file parser, we dont need it

PR #681, opened 2025-10-02
Fix duplicate input actions in Playwright test generation

thehimmat:
## Problem
Staktrak generated duplicate `page.fill()` calls for the same input field with the same value. This was caused by a race condition between the debounce timer and blur event handler - both were pushing "complete" input actions independently.

## Solution
1. **Blur handler fix**: Modified to only push action when debounce timer exists (user blurred before timer fired). If timer doesn't exist, the action was already captured by the debounce completion.

2. **Input deduplication**: Added safety check to remove consecutive duplicate input actions on the same element with the same value, following the existing waitForUrl deduplication pattern.

3. **Build format change**: Changed playwright-generator from ESM to IIFE format to properly expose `window.PlaywrightGenerator` global.

## Testing
Verified fix eliminates duplicates in both localhost and production environments. All TypeScript compilation successful, no conflicts with main branch.

## Files Changed
- `src/index.ts` - Blur handler conditional logic fix
- `src/actionModel.ts` - Input action deduplication
- `src/playwright-generator.ts` - Extracted generatePlaywrightTest as named export
- `package.json` - Changed build format from ESM to IIFE for proper window global
- `dist/playwright-generator.js` - Large diff due to IIFE wrapping (esbuild format change, minimal logic changes)
- `dist/staktrak.js` - Rebuilt with input deduplication changes

### DIFF
--- a/mcp/tests/hooks.js
+++ b/mcp/tests/hooks.js
@@ -76,6 +76,16 @@ export function useIframeMessaging(iframeRef, initialURL) {
             console.log("Staktrak setup message received");
             break;
           case "staktrak-results":
+            // TODO: RecordingManager data sync issue
+            // RecordingManager's trackingData comes from real-time staktrak-event messages,
+            // but staktrak-results includes post-processing (e.g., filterClickDetails removes
+            // clicks associated with assertions). This means generated tests may have extra
+            // click actions that should have been filtered.
+            //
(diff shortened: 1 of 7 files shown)


thehimmat:
I'm doing a bit more testing and code cleanup, so changing this back to draft status for a bit

PR #688, opened 2025-10-04
fix: test calls class edge 

fayekelmith:
- [x] Resolve test calls class edge creation
- [x] Fix querying for test calls class


Evanfeenstra (approved):
looks good

make sure its well tested, can there be more in ast/src/testing/ruby/mod.rs?

PR #754, opened 2025-10-18
fix: multi regex & test_filters

fayekelmith:
- [X] accept multi regexes
- [X] per Test type


ast/src/lang/graphs/mod.rs
```diff
@@ -105,6 +105,15 @@ impl NodeRef {
     }
 }
 
+#[derive(Debug, Clone, Default)]
+pub struct TestFilters {
```
Evanfeenstra: ignore_dirs should be in here too

PR #802, opened 2025-11-08
fix: setup testing and add node finder from import workflow

### LINKED ISSUE #801: func_target_file_finder
in func_target_file_finder there are 3 steps:

- find_only_one_function_file
- find_function_in_same_file
- find_function_in_same_directory

but we need another step: find the target based on the import in the file.

so given this test file
```
import { describe, it, expect } from 'vitest'
import { convertSatsToUSD } from '@/lib/utils'

describe('Happy Path - Valid Conversions', () => {
    it('should convert small amounts correctly', () => {
      // 1 sat at $50,000/BTC = $0.0005
      expect(convertSatsToUSD(1, 50000)).toBe('0.00')
  })
})
```

convertSatsToUSD is not found (because there are 2 separate functions with the same name in different files)

So we need to get the Import section from the `current_file`, and parse that to find which file the convertSatsToUSD was imported from. Then normalize to remove the `@/`. Then find convertSatsToUSD function in with find_node_by_name_and_file_end_with

fayekelmith:
- Closes #801

### DIFF
--- a/ast/src/testing/nextjs/mod.rs
+++ b/ast/src/testing/nextjs/mod.rs
@@ -109,7 +109,7 @@ pub async fn test_nextjs_generic<G: Graph>() -> Result<()> {
     let functions = graph.find_nodes_by_type(NodeType::Function);
     nodes += functions.len();
     if use_lsp {
-        assert_eq!(functions.len(), 45, "Expected 45 Function nodes with LSP");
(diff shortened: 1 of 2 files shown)


Evanfeenstra:
can you add something like “clean_resolved_path” for Stack trait

for react it would trim "@/" and also ".js"

since for some typescript projects, they import from ".js" even through its a ".ts" file

PR #805, opened 2025-11-10
fix: react hook patterns coverage and test patterns

fayekelmith:
- [X] Mock patterns on server
- [x] Outline desired connections
- [x] Implement the capture route

### DIFF
--- a/ast/src/lang/parse/format.rs
+++ b/ast/src/lang/parse/format.rs
@@ -1016,6 +1016,17 @@ impl Lang {
                     called, &tf.file
                 ));
                 fc.target = tf.into();
+            } else if let Some(ref operand) = fc.operand {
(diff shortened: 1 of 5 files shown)


ast/src/lang/call_finder.rs
```diff
@@ -66,6 +67,16 @@ pub fn func_target_file_finder<G: Graph>(
         return Some(tf);
     }
 
+    // Fifth try: If operand exists, try operand-based resolution
+    if let Some(ref operand) = operand {
+        if let Some(target_file) = find_function_with_operand(operand, func_name, graph) {
```
Evanfeenstra: special case for typescript: if the fn starts with `use`, then search for the OPERAND directly as a function 

that will create a CALLS edge from the test or component to the hook

(this is for tests only???)

PR #824, opened 2025-11-15
refactor: eliminate redundant test/function harvesting with mem tracking

### LINKED ISSUE #821: simplify test parsing
we have `collect_functions_and_tests`... but we also have separate test queries

we need to instead, 
- parse a file, both tests (classified) and functions
- using 2 separate treesitter queries
- if a `Function` already was marked as a Test, then dont add it as a function
- use a memory reference of the file?

fayekelmith:
Closes #821

### DIFF
--- a/ast/src/lang/mod.rs
+++ b/ast/src/lang/mod.rs
@@ -20,6 +20,7 @@ use std::str::FromStr;
 use streaming_iterator::{IntoStreamingIterator, StreamingIterator};
 use tracing::trace;
 use tree_sitter::{Node as TreeNode, Query, QueryCursor};
+use std::collections::HashSet;
 
 pub struct Lang {
     pub kind: Language,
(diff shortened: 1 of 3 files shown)


ast/src/lang/mod.rs
```diff
@@ -416,93 +444,49 @@ impl Lang {
         graph: &G,
         lsp_tx: &Option<CmdSender>,
     ) -> Result<(Vec<Function>, Vec<TestRecord>)> {
-        let qo = self.q(&self.lang.function_definition_query(), &NodeType::Function);
-        let mut funcs1 = self.collect_functions(&qo, code, file, graph, lsp_tx)?;
-        self.attach_function_comments(code, &mut funcs1)?;
-        let (funcs, filtered_tests) = self.lang.filter_tests(funcs1);
         let mut tests: Vec<TestRecord> = Vec::new();
         let mut seen_tests = std::collections::HashSet::new();
+        let mut identified_tests: std::collections::HashSet<(String, String, usize)> = std::collections::HashSet::new();
```
Evanfeenstra: any way we can simplify this? like using BtreeSet with `NodeKeys` instead of (String, String, usize)?

fayekelmith: HashSet is faster though, O(1), we can use BTreeSet with NodeKeys but then it'll be slightly slower O(logn)... not super significant, but ...

PR #937, opened 2026-01-13
feat: monorepo resolution

fayekelmith:


ast/src/lang/graphs/neo4j/graph.rs
```diff
@@ -124,6 +124,13 @@ impl Neo4jGraph {
             .context("Neo4j Connection is not established")
     }
 
+    pub async fn clear_graph(&self) -> Result<()> {
```
Evanfeenstra: never clear the whole graph

PR #1006, opened 2026-02-11
refactor: build graphs in parrallel using rayon

fayekelmith:
Tested with Sphinx-bounties site: 
- Ingestion time Before: around 26 minutes: 1626s
- ingestion time After: around 17 minutes: 1035s
around 36% increase in speed.


Evanfeenstra:
Why only use parallel if using Neo4jGraph?

PR #1154, opened 2026-04-03
fix: update descriptions to skip built-in nodes

fayekelmith:


Evanfeenstra:
functions like 
```ts
json: async () => ({
          success: true,
          message: "Push successful",
          prs: { "test-org/test-repo": "https://github.com/test-org/test-repo/pull/123" },
        }),
        ```
        
        are not "built in" ... the issue is they are nested.
        
 Nested function/datamodels etc shouldnt be described

Evanfeenstra (commented):
the right way to handle this is avoid describing nest functions/datamodels etc

PR #1166, opened 2026-04-08
fix: prevent no-code ndoes from being returned from the graph

### LINKED ISSUE #1165: filter stakgraph_search
only code nodes. no Hint nodes

fayekelmith:
Closes #1165

### DIFF
--- a/mcp/src/graph/graph.ts
+++ b/mcp/src/graph/graph.ts
@@ -9,6 +9,7 @@ import {
   NodeType,
   EdgeType,
   GraphResponse,
+  code_node_types,
 } from "./types.js";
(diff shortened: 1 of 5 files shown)


mcp/src/graph/graph.ts
```diff
@@ -80,8 +80,8 @@ export async function search(
   sortBy: SearchSortBy = "relevance"
 ) {
   const skip_node_types: NodeType[] = tests
-    ? []
-    : (["UnitTest", "IntegrationTest", "E2etest"] as NodeType[]);
+    ? (["Hint", "Prompt", "PullRequest", "Mock", "Learning", "Scope"] as NodeType[])
```
Evanfeenstra: only source code nodes for stakgraph_search. We can have any number of node types in production graphs

PR #1167, opened 2026-04-08
fix: clean out nested-in nodes that create noise

fayekelmith:
- Clean at the end strategy


Evanfeenstra:
- capture EVERY nested_in edge (any node type can be nested in any node type)
- process the graph like normal
- delete any node that is nested that does not have a CALLS|HANDLER|RENDER

PR #1186, opened 2026-04-17
feat: update sessions for all agentic workflows

fayekelmith:
Screenshot 2026-04-21 at 12 52 41 PM

### DIFF
--- a/mcp/src/repo/session.ts
+++ b/mcp/src/repo/session.ts
@@ -13,6 +13,8 @@ import path from "path";
 
 const SESSIONS_DIR = process.env.SESSIONS_DIR || ".sessions";
 
+const sessionMeta = new Map<string, { source: string; start_time: string }>();
+
 export interface Session {
   id: string;
   messages: ModelMessage[];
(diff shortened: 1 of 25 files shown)


mcp/src/repo/session.ts
```diff
@@ -42,18 +42,30 @@ function getSessionFile(sessionId: string): string {
...
+export function createSession(id?: string, system?: string, source?: string): string {
   const sessionId = id || randomUUID();
   const filePath = getSessionFile(sessionId);
+  const meta = { role: "metadata", source: source || "unknown", start_time: new Date().toISOString() };
+  appendFileSync(filePath, JSON.stringify(meta) + "\n");
   if (system) {
     const systemMsg: ModelMessage = { role: "system", content: system };
     appendFileSync(filePath, JSON.stringify(systemMsg) + "\n");
-  } else {
-    appendFileSync(filePath, "");
```
Evanfeenstra, line 52: why delete this?

Evanfeenstra:
- SESSION_MAX_AGE_MS should be configurable via env, default to 30 days
- no role:metadata message... instead, store the session info in neo4j (AgentSession node type with properties for each metric, make sure they are numerical)
- model, time (length), start, end, total_tokens, input_tokens, output_tokens, repo(s), and of course session id
- we need ONE cohesive agent for gitree_process, even though different calls DO NOT carry the actual session context
  - for example: LLM decides which features the PR is part of. Then LLM summarizes changed features. These are separate calls, that DO NOT share context totally... BUT we do want to trace them together

Evanfeenstra:
add nonCumulative arg to get_context so that we can NOT add in all previous messages for certain calls (such as applyPrDecision->applyDecisionToFeature->exploreNewFeature) or the clue analysis

line 306 of agent.ts

```ts
  if (inputSessionId) {
    if (sessionExists(inputSessionId)) {
      if (nonCumulative) {
        sessionId = inputSessionId;
        previousMessages = [];
      } else {
        sessionId = inputSessionId;
        previousMessages = loadSessionMessages(sessionId);
      }
    } else {
      sessionId = createNewSession(inputSessionId, instructions);
    }
  }
```

fayekelmith:
> add nonCumulative arg to get_context so that we can NOT add in all previous messages for certain calls (such as applyPrDecision->applyDecisionToFeature->exploreNewFeature) or the clue analysis

- clueAnalysis and exploreNewFeature, because they use `get_context` they use the session_id to load messages from file which can pollute the context.

Evanfeenstra:
right now you only write token counts if the repo/agent is called for the first time

from hive, there are multiiple rounds of repo/agent with the same session id (each message is a new repo/agent, building on top of the same session)

so you need to test: does loading an existing session automatically accumulate the input and output tokens? if so, we can just overwrite each time. if not, we need to manually accumulate

@fayekelmith

Evanfeenstra:
subAgents run on different servers with different filesystems... make the UI shows correctly (parent can store the token usage in the db, but shouldn't try to make fake jsonl session, and shouldnt try to load the child sessoin from the UI, since its not actually available)

U can change SESSIONS_DIR for one stakgraph to test

PR #1191, opened 2026-04-20
Generated with Hive: Add sidebar node hover and selection highlight with isolated NodeHoverHighlight component

Rassl:
Add sidebar node hover and selection highlight with isolated NodeHoverHighlight component


Rassl:
wrong repo

PR #1228, opened 2026-05-01
feat: stakgraph search include and exclude

### LINKED ISSUE #1222: stakgraph_search filters
filters on the search to both include and exclude.

support blobs like

include:
- `**/*.ts` (any ts file)
- `stakwork/hive/**` (only in this repo)

exclude:
- `__tests__, dist` (nothing in tests dir)

fayekelmith:
CLoses #1222

### DIFF
--- a/mcp/src/graph/graph.ts
+++ b/mcp/src/graph/graph.ts
@@ -94,7 +94,9 @@ export async function search(
   output: OutputFormat = "snippet",
   skip_node_types: NodeType[] = [],
   language?: string,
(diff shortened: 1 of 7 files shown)


Evanfeenstra:
did you test this carefully?

fayekelmith:
> did you test this carefully?

yes it works.... but because neo4j doesn't support blob, had to use CONTAINS, STARTS WITH, etc. and things like .tsx might get into filters that are for .ts .... but generally it works.

PR #1236, opened 2026-05-06
feat: implement default cache on all agentic flows

fayekelmith:


mcp/src/tools/explore/tool.ts
```diff
@@ -168,26 +172,34 @@ export async function get_context_explore(
...
+    : Array.isArray(cachedSystem)
+    ? cachedSystem
+    : [cachedSystem];
+  const promptMessages: ModelMessage[] = typeof prompt === "string"
+    ? [{ role: "user", content: prompt }]
+    : prompt;
   const { steps, totalUsage } = await generateText({
     model,
     tools,
-    prompt,
```
Evanfeenstra, line 174: why remove prompt/system and use `messages` instead?

Evanfeenstra: I see. We dont need cacheControl on every message and tool. Using auto caching is best https://platform.claude.com/docs/en/build-with-claude/prompt-caching#automatic-caching

PR #1250, opened 2026-05-10
fix: probe and fix stagkraph search

fayekelmith:
**Changes:**
- Score floor on vector results (drops weak matches)
- Multi-word fulltext queries use OR + phrase boost instead of AND
- Lucene special-char escape fix (prevents crashes on paths like `/api/[id]`)
- Vector over-fetch (`limit × 3`) for better RRF candidate pool
- RRF overlap bonus for results appearing in both vector + fulltext
- Fulltext index includes `description` field

## Benchmark Results (138 queries, sphinx-bounties)

| Metric | `main` (before) | fix branch (after) | Δ |
|--------|-----------------|-------------------|---|
| Hit@1 | 85% (118/138) | 89% (124/138) | +6 |
| Hit@3 | 86% (120/138) | 97% (135/138) | +15 |
| Hit@5 | 86% (120/138) | 99% (137/138) | +17 |

| Category | main @1 / @3 | fix @1 / @3 | Δ@1 |
|----------|-------------|------------|-----|
| exact-fn (30) | 30/30 / 30/30 | 30/30 / 30/30 | — |
| endpoint (26) | 10/26 / 10/26 | 26/26 / 26/26 | **+16** |
| datamodel (25) | 25/25 / 25/25 | 25/25 / 25/25 | — |
| semantic (57) | 53/57 / 55/57 | 43/57 / 54/57 | -10 @1, +10 @3 |

**18 improvements, 2 regressions.**
The endpoint category was completely broken on `main` — all 16 paths containing `[id]`, `[pubkey]`, or `[paymentHash]` returned no results due to a Lucene escape bug. All fixed.
The 2 regressions are edge-case implicit-concept semantic queries.

### DIFF
--- a/mcp/src/graph/neo4j.ts
+++ b/mcp/src/graph/neo4j.ts
@@ -1780,75 +1780,51 @@ async function execute_batch(session: Session, batch: MergeQuery[]) {
 /**
  * Prepares a fulltext search query for Neo4j by properly handling special characters
  */
+// Returns true if a raw (unescaped) term is safe to use in Lucene wildcard queries.
+// Wildcard queries do not honour escape sequences, so terms containing special
+// Lucene characters must never be used in wildcard position.
+function isSimpleTerm(raw: string): boolean {
+  return /^[\w\-.]+$/.test(raw);
+}
+
 export function prepareFulltextSearchQuery(searchTerm: string): string {
   const words = searchTerm.trim().split(/\s+/).filter(Boolean);
 
   if (words.length === 1) {
-    const word = escapeSearchTerm(words[0]);
+    const raw = words[0];
+    const word = escapeSearchTerm(raw);
+    const simple = isSimpleTerm(raw);
     const fieldMatches = [
       `name:${word}^10`,
-      `file:*${word}*^4`,
       `body:${word}^3`,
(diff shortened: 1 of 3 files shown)


Evanfeenstra:
you added `description` to the index but not the prepareFulltextSearchQuery

mcp/src/graph/neo4j.ts
```diff
@@ -1780,75 +1779,52 @@ async function execute_batch(session: Session, batch: MergeQuery[]) {
...
 export function prepareFulltextSearchQuery(searchTerm: string): string {
   const words = searchTerm.trim().split(/\s+/).filter(Boolean);
 
   if (words.length === 1) {
-    const word = escapeSearchTerm(words[0]);
+    const raw = words[0];
+    const word = escapeSearchTerm(raw);
+    const simple = isSimpleTerm(raw);
     const fieldMatches = [
       `name:${word}^10`,
```
Evanfeenstra, line 1789: why remove file? is it not needed? sometimes an agent might search for filename...

fayekelmith: It's removed only for multi-word queries, because filenames don't match them, right? I can add them, but it would just be noise...

PR #1260, opened 2026-05-15
feat: overflow context summarizer

fayekelmith:


Evanfeenstra:
This is a good idea. We need something like this, but I think it's risky to delete messages because the user can refer to an earlier message that they sent. Thats why the current truncateOldToolResults only removes tool results (which can always be re-called)

PR #1276, opened 2026-05-18
Generated with Hive: Add timeouts and abort handling to git operations and busy tracker

gonzaloaune:
Add timeouts and abort handling to git operations and busy tracker


fayekelmith:
The happy path works (cloning public and private repos), but manual abort doesn't kill the underlying git process — stream stays open the full 10 min until the timeout fires...

Evanfeenstra:
> The happy path works (cloning public and private repos), but manual abort doesn't kill the underlying git process — stream stays open the full 10 min until the timeout fires...

is there a good way to fix this @fayekelmith

fayekelmith:
> > The happy path works (cloning public and private repos), but manual abort doesn't kill the underlying git process — stream stays open the full 10 min until the timeout fires...
> 
> is there a good way to fix this @fayekelmith

I've wired the AbortSignal to `simple-git's` abort option, so it terminates any subprocess

PR #1281, opened 2026-05-19
fix: graph querying optimization

fayekelmith:


ast/src/lang/graphs/neo4j/operations/migration.rs
```diff
@@ -475,7 +475,7 @@ pub fn clear_existing_graph_query(root: &str) -> (String, BoltMap) {
 
 pub fn set_missing_data_bank_query() -> String {
     r#"
-        MATCH (n)
+        MATCH (n:Data_Bank)
```
Evanfeenstra, line 478: MATCH (n:Data_Bank)
WHERE n.Data_Bank IS NULL

huh??? is that even possible?

fayekelmith: it's setting the `property` not the `label` `n.Data_Bank` not `n:Data_Bank`

PR #1345, opened 2026-06-10
Generated with Hive: Fix default_do_lsp to return correct LSP support based on USE_LSP env var

tomsmith8:


fayekelmith:
ideally, the tests added to the work should live somewhere else... in a module in `ast/src/testing/*`

PR #1373, opened 2026-06-16
feat: type extraction for ts only... for now.

fayekelmith:


ast/src/lang/parse/format.rs
```diff
@@ -933,6 +934,23 @@ impl Lang {
             }
         }
 
+        if matches!(self.kind, Language::Typescript) {
```
fayekelmith, line 937: Do we need this? Consider making this generic later.

PR #1494, opened 2026-07-27
feat: add summarize param to GET /repo/agent/session

fayekelmith:


Evanfeenstra:
Can this go throught the normal ToolLoopAgent pipeline AS IF its a regular message in the conversation? so we continue to get the 10x cheaper kv cache pricing. But DONT actually add the msg onto the context history

PR #1521, opened 2026-08-05
Recover agent runs that stall without delivering an answer

Evanfeenstra:
## Problem

Long agentic runs (especially Opus and Kimi K3) intermittently "just stop" without delivering what they promised — sometimes the last thing in the transcript is a reasoning block, sometimes a tool call, and the run still reports success with a garbage answer scraped from leftover narration.

Investigation on this branch identified **three distinct stall mechanisms**, all now handled:

1. **Voluntary stall** (`end_turn`): the model narrates a plan ("Let me now…") or emits only reasoning and ends its turn without tool calls or `[END_OF_ANSWER]`.
2. **Output-token truncation** (`length`): a step hits the per-call output cap mid-answer or mid-thinking (thinking tokens count against the same budget).
3. **Mid-stream connection error** (`error`): a transient network failure to the provider (reproduced live: `read ETIMEDOUT` to OpenRouter/Moonshot) kills the stream right after a reasoning block; the AI SDK surfaces it as a final step with `finishReason: "error"` instead of throwing, so the loop exits "successfully". This is the "reasoning block is the last thing I see" symptom.

## Fix: continuation machinery

When a run ends without a proper termination (no `[END_OF_ANSWER]`, no `ask_clarifying_questions`, no pending tool calls), the full conversation plus a synthetic user nudge is re-sent, resuming from the exact point of interruption (prompt caching makes this cheap):

- **Per-kind nudge wording** — stalls are told to execute their plan now (with an explicit escape hatch to ask the user a question instead of fabricating an answer); truncations are told to continue from the exact cutoff without repeating (and the nudge states the actual token limit); connection errors are told the interruption was transient and to re-issue the unexecuted tool call.
- **Per-kind budgets** — voluntary stalls get 2 nudges; involuntary interruptions (truncation/error) share a progress-gated allowance of 5, so a large text deliverable can span up to 6 stitched segments (~6× the per-call cap). A continuation that produces zero output tokens stops the loop.
- Nudges are persisted tagged (`providerOptions.stakgraph.continuationNudge`) so transcript consumers and turn counting can distinguish them from real user turns.
- Detection uses the raw provider stop reason (recorded per step as `rawFinishReason` for observability): Anthropic's `stop_sequence` vs `end_turn` distinguishes a proper finish from a stall. OpenAI-compatible providers report `stop` either way, so their voluntary stalls are conservatively left alone (no false positives); their truncations and connection errors are still caught.

## Supporting fixes

- **One-shot 400 retry** resuming from the in-flight step (the SDK only auto-retries 408/409/429/5xx), plus API error response bodies folded into error messages.
- **Provider-aware output cap**: default 128k for Anthropic (provider clamps known models to their true max), 64k for OpenAI-compatible hosts; `MAX_OUTPUT_TOKENS` env still overrides.
- **Time-budget nudges** at 50% / 75% / 92% of `BUSY_TIMEOUT_MINUTES` (default 120), injected between steps, so runs land before the hard-kill abort discards their work.
- **Kimi context limits corrected**: K3 is 1M (was falling back to 128k and truncating tool results at ~115k); K2.x are 262k, not 128k. Verified against the OpenRouter catalog.
- **Kimi routing pinned to Moonshot** via OpenRouter provider preferences (only Moonshot's endpoint has prompt caching and reliable tool-call parsing), with cached-token counts folded from provider metadata into step usage.
- **`generate_docx` accepts `markdownPath`**: large documents can be built incrementally with bash appends and then converted — a large inline `markdown` argument must fit one output-capped message and reliably failed on Kimi. The tool description steers large documents to this flow.

## Validation (live runs against harvey-labs prod task)

| Scenario | Result |
|---|---|
| Opus, clean run | 55k-token answer, `stop_sequence` finish, zero false nudges |
| Opus, forced 8k cap (text answer) | 5 resume continuations, mid-word stitching, zero duplication |
| Opus, forced 8k cap (file deliverables) | 1 truncation → nudge → model switched to chunked writes → both .docx produced |
| Kimi K3, wifi drop mid-run | Reproduced the silent-failure mechanism that motivated the `error` continuation |
| Kimi K3, prod config | Both deliverables via `markdownPath`, 36 steps, zero nudges needed, 1.2M cached tokens |

Unit tests cover `needsContinuation` (all three kinds plus proper terminations), nudge tagging, time-budget thresholds, and `markdownPath` (58 docgen + 54 utils tests passing).

### DIFF
--- a/mcp/package.json
+++ b/mcp/package.json
@@ -59,7 +59,7 @@
     "@noble/ed25519": "2.3.0",
     "@noble/hashes": "1.7.1",
     "@octokit/rest": "22.0.1",
-    "@openrouter/ai-sdk-provider": "6.0.0-alpha.1",
+    "@openrouter/ai-sdk-provider": "2.10.0",
     "@playwright/test": "1.58.1",
     "ai": "6.0.175",
     "archy": "1.0.0",
--- a/mcp/src/aieo/package.json
+++ b/mcp/src/aieo/package.json
@@ -25,10 +25,10 @@
   "license": "ISC",
   "description": "typescript ai utils",
   "dependencies": {
-    "@ai-sdk/anthropic": "3.0.92",
+    "@ai-sdk/anthropic": "3.0.105",
     "@ai-sdk/google": "3.0.67",
     "@ai-sdk/openai": "3.0.61",
-    "@openrouter/ai-sdk-provider": "6.0.0-alpha.1",
+    "@openrouter/ai-sdk-provider": "2.10.0",
     "ai": "6.0.175",
     "zod": "4.3.6"
   },
--- a/mcp/src/aieo/src/provider.ts
+++ b/mcp/src/aieo/src/provider.ts
@@ -522,8 +522,14 @@ const MODEL_CONTEXT_LIMITS: Record<string, number> = {
   // OpenAI
   "gpt-5": 128_000,
   "gpt-4.1-mini": 1_000_000,
-  // OpenRouter
-  "moonshotai/kimi-k2.6": 128_000,
+  // OpenRouter — values from the OpenRouter model catalog
+  // (https://openrouter.ai/api/v1/models, context_length).
+  "moonshotai/kimi-k3": 1_048_576,
+  "moonshotai/kimi-k2.7-code": 262_144,
+  "moonshotai/kimi-k2.6": 262_144,
+  "moonshotai/kimi-k2.5": 262_144,
+  "moonshotai/kimi-k2-thinking": 262_144,
+  "moonshotai/kimi-k2": 131_072,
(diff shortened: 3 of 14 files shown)


Evanfeenstra:
Added a second commit: Kimi/Moonshot runs are now pinned (with fallbacks) to Moonshot's own OpenRouter endpoint — the only kimi-k3 host with automatic prompt caching. Verified directly against the OpenRouter API: repeat call reports `cached_tokens: 1280/1466` and ~67% lower cost. Note the `usage`/`provider` options had to be passed as model-construction options; the call-level `providerOptions.openrouter` channel is ignored by the v6 alpha provider.

PR #1531, opened 2026-08-06
Harden gitree date parsing so malformed Concept nodes can't 500 the concepts list

Evanfeenstra:
## Problem

The openlaw workspace's /learn Concepts list was empty. Root cause on the swarm:

```
Error listing concepts: RangeError: Invalid time value
    at Date.toISOString (<anonymous>)
    at listConcepts (file:///usr/src/app/build/gitree/service.js:7:28)
```

Concept nodes created outside the normal `saveConcept` path can lack the epoch-seconds `date` property (or store it in another format). `nodeToConcept` did `new Date(numberOrUndefined(props.date)! * 1000)`, yielding an Invalid Date; `listConcepts` then threw on `.toISOString()`, so **one malformed node emptied the entire concepts list for the workspace**.

## Fix

- **`safeDate()` helper**: accepts epoch seconds, epoch millis, neo4j `Integer`s, numeric strings, ISO strings, and neo4j temporal types (seconds vs millis disambiguated by magnitude). Missing/unparseable values fall back to epoch 0 — never an Invalid Date, so `toISOString()` can't throw downstream.
- Applied at all node→record date conversions: `nodeToConcept` (`createdAt`, `lastUpdated`, `cluesLastAnalyzedAt`), `nodeToPR.mergedAt`, `nodeToCommit.committedAt`, `nodeToClue` (`createdAt`/`updatedAt`).
- **`getAllConcepts` skip-and-warn**: a node that still fails conversion for any other reason (e.g. malformed JSON in `newDeclarations`) is dropped with a `console.warn` naming its id, instead of failing the whole listing.

Nodes with bad dates are **kept** (sorted last with a 1970 timestamp), not skipped; only nodes that throw for other reasons are skipped, and those are logged per request so they're visible in swarm logs.

## Verification

- `tsc --noEmit` clean.
- `safeDate` behavior checked across shapes: missing/null/garbage/empty → epoch 0; epoch seconds, epoch millis, neo4j Integer, numeric string, ISO string all parse to the correct instant.

### DIFF
--- a/mcp/src/gitree/store/graphStorage.ts
+++ b/mcp/src/gitree/store/graphStorage.ts
@@ -201,6 +201,7 @@ export class GraphStorage extends Storage {
       const result = await session.run(`
         MATCH (c:Concept)
         WHERE c.embeddings IS NULL
+          AND c.id IS NOT NULL
           AND (
             (c.name IS NOT NULL AND c.name <> "") OR
             (c.description IS NOT NULL AND c.description <> "")
@@ -653,10 +654,12 @@ export class GraphStorage extends Storage {
       
       const result = await session.run(
         `
-        MATCH (f:Concept {id: $id})
+        MATCH (f:Concept)
+        WHERE f.id = $id OR f.node_key = $rawId OR f.ref_id = $rawId
         RETURN f
+        LIMIT 1
         `,
-        { id: fullId }
+        { id: fullId, rawId: id }
       );
 
       if (result.records.length === 0) {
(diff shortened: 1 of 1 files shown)


Evanfeenstra:
Pushed a second commit after tracing where the malformed concepts come from: Jarvis-created Concept nodes (`POST /v2/nodes`, e.g. the `create_triplet` tool / hive lingo routes) carry `node_key`/`ref_id`/`date_added_to_graph` but no `id` or `date` property.

Beyond the date crash, the missing `id` caused two more problems, now handled:

- `backfillConceptEmbeddings` runs on every `initialize()` and writes back via `MATCH (c:Concept {id: $id})` — with `id = null` the SET never matched, so id-less concepts were re-embedded (an LLM call each) on **every** `/gitree/concepts` request. The backfill now skips them.
- List consumers received `id: null` (hive's sidebar groups by `id.split("/")` and would throw). `nodeToConcept` now falls back to `node_key`/`Data_Bank`/`ref_id`, and `getConcept` matches those too so the fallback ids resolve on detail lookups.

PR #1559, opened 2026-08-14
Generated with Hive: Return envelope with returned, truncated, total_edges, and neighbors from graph_neighbors

gonzaloaune:


gonzaloaune:
not used for now

PR #1568, opened 2026-08-17
Live Turn-node emission: mirror agent sessions into the graph as they run

Evanfeenstra:
## What

Emits `(AgentSession)-[:HAS_TURN]->(Turn)-[:NEXT]->(Turn)…` chains into Neo4j **live, after every agent step**, for both the repo agent (`agent.ts`) and `graph_sub_agent` children (`toolsJarvis.ts`) — plus live `(Turn)-[:READ_CONCEPT]->(Concept)` edges the moment a tool result hands back a Concept body.

## Why

A post-hoc workflow (build_trace_edges.py) already builds Turn chains, but inspection of prod showed **zero `HAS_TURN` edges exist** — all ~1,600 chains are orphans, joinable to their AgentSession only by string-parsing `turn_id` prefixes. And since chains were only built after the fact, there was nothing to watch while a session runs. This makes the session graph grow in real time: session node appears at start, turns stream in per step, concept edges land as they're read.

## Node/edge shape

Matches the prod backfill shape so live and backfilled chains are indistinguishable (verified against prod data):

- `turn_id` = `{agent}-{sessionId}-turn-{order}`; `node_key` = `"turn-"` + turn_id stripped of non-alphanumerics (prod: `repair-agent-147813394-turn-0` → `turn-repairagent147813394turn0`)
- labels `:Data_Bank:Turn:Node`; one Turn per content part (`user_input` / `reasoning` / `tool_call` / `tool_result`); tool results truncated to 100 chars + `"..."`; `NEXT {weight:1}` chain; final reasoning turn retyped to `response` at session end
- Additive props the backfill lacks: `timestamp` (epoch ms, when the moment happened) and `tool_call_id` (pairs call/result turns even when parallel calls interleave)
- Deliberate deviations (all data improvements): real tool names on `tool_result` turns (prod has `"unknown"`), no `unknown`-type turns for system messages, continuation nudges and empty thinking blocks skipped, one `response` per run on multi-run sessions

## Key pieces

- **`repo/turns.ts` (new)** — shared emitter: classification, per-session order counter (persisted in a `.turns.json` sidecar so resumed sessions continue their chain across processes), concept detection, serialized fire-and-forget writes. Self-healing: both NEXT endpoints are MERGEd by deterministic node_key, so a lost batch re-links on the next one.
- **AgentSession stub at `createSession()`** — node now exists from run start with `status: 'running'` (crashed sessions stay visible instead of never existing), giving turn 0 its `HAS_TURN` anchor immediately. `appendSessionEnd` finalizes the same node; `model`/`provider` moved into the upsert's `SET` clause (empty-guarded) since `ON CREATE` no longer fires for stubbed sessions.
- **`(parent)-[:SPAWNED]->(child)` edge** — parent/child sessions were previously linked only by the `parent_session_id` property; the stub creates a real traversable edge (property kept, existing queries untouched).

## Notes for reviewers

- The graph is an index, never the record: session JSONL stays authoritative, every write is fire-and-forget and logged-not-thrown, mirroring the existing `syncReflectionEdges` posture.
- `text_embeddings` on Turn nodes are jarvis-ingest artifacts; direct writes don't populate them. If anything searches Turns semantically, that needs a jarvis-side backfill.
- Sessions API passes `status` through as an opaque string, so the new `running` value flows through cleanly.
- Tests: 8 new unit tests in `turns.test.ts` (node_key parity with prod, ordering/chaining, truncation, concept links, resume, finalize); full `test:node` suite green (533), `tsc --noEmit` clean.

Verify after deploy:
```cypher
MATCH (s:AgentSession {status:'running'})-[:HAS_TURN]->(t0)
MATCH (t0)-[:NEXT*0..]->(t) RETURN s.node_key, t.order, t.turn_type, t.tool ORDER BY t.order;
```


Evanfeenstra:
Added in 65c2a66: **startup backfill for pre-live-emission sessions** (no SSH / manual ops — runs from the boot task block like the concept backfills, marker-guarded so it's once-only per instance).

- `turnsFromTranscript()` in turns.ts: pure classifier from a persisted transcript to the exact chain live emission would produce — same deterministic node_keys, so backfilled and live chains can't diverge. Retypes the last reasoning turn to `response` pre-write (python parity), skips system messages and continuation nudges, detects Concept reads.
- `turnBackfill.ts`: oldest-first sweep with mtime high-water marker; skips sessions that have a live turn cursor; 10-min recency guard against racing a just-started run; never writes orphan chains (no AgentSession node → zero turns written by design); backfilled turns carry **no** timestamp rather than a fake one. Writes the cursor sidecar afterward so a later live resume continues the chain.
- `UPSERT_TURNS_QUERY` now bumps `turn_count` / `last_turn_at` on the AgentSession node in the same batch (monotonic, for session-list UIs).

Tests: 538 green (13 in the turn suite), tsc clean.

PR #1631, opened 2026-09-02
feat: Auditor agent — pod-resident verification agent (MVP)

fayekelmith:


fayekelmith:
Superseded by the portable-verification approach (#1640). The bespoke pod-resident agent's core — the proof-guarded submit_verdict, the verdict schema, and the probe tools (network/console/db/http/timing) — now lives in the shared MCP tool layer, so any generic agent (in-process, or goose over MCP) can verify instead of a dedicated binary. Validated end-to-end via the hive trigger → shared tools → callback → verdict-card loop. Closing in favor of #1640; the hive surfacing (#5217) and the pod door (staklink #125) remain.

PR #1653, opened 2026-09-04
lab: wfbench — validation as judge evidence, namespace as run input

Evanfeenstra:
Two small pieces toward the Hive "run in vein" toggle.

- **Validation as evidence.** `wfbench-run` now runs `meta/validate-workflow` on the produced YAML and `wfbench/build-materials` inlines the result as a `VALIDATION` material (`ok`, `errors`, `warnings`, `summary`). Structural criteria ("workflow is structurally valid") get judged on the engine's own check plus the rerun status instead of on the judge's reading of the YAML. Not a gate: a candidate that validates poorly is still launched and judged.
- **Namespace as input.** `input.namespace || params.namespace` on every roster/record write. Hive already upserts the EvalSet/EvalRequirement roster through jarvis before dispatch; passing that partition lets vein's writes merge onto the same nodes by id rather than creating a second roster under `wfbench`.

Verification: mcp `tsc` clean; `npx tsx src/lab/wfbench/smoke.ts` green (both workflows static-validate, build-materials asserts the VALIDATION section).

### DIFF
--- a/mcp/src/lab/AGENTS.md
+++ b/mcp/src/lab/AGENTS.md
@@ -615,7 +615,9 @@ workflow (design + step map: `plans/wfbench-harness.md`). One task in
 rerun_expected_output?, webhook_url? }` — Hive's payload shape), one callback
 out. `wfbench-run`: graph roster (EvalSet → EvalRequirement×N, EvalTrigger,
 HAS_BASELINE_TRIGGER on the EvalSet's first trigger else HAS_TRIGGER — 58313's
-ids, written with vein's `graph/*` steps under `params.namespace`) ‖ the
+ids, written with vein's `graph/*` steps under `input.namespace ||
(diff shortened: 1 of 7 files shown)


Evanfeenstra:
Live run 1788554443025 on this branch with `input.namespace: wfbench-ns-test`: 4/4, candidate v4. Namespace registered and the whole roster + record chain landed under `wfbench-ns-test` (1 EvalSet, 4 EvalRequirement, 1 EvalTrigger on HAS_BASELINE_TRIGGER since that partition had none, 1 EvalTriggerOutput, 4 CriterionResult) — the `wfbench` partition from earlier runs untouched. The judge materials carried `### VALIDATION: static validation: ok (0 error(s), 0 warning(s))` from `meta/validate-workflow`.

PR #1670, opened 2026-09-09
strut: consume as a git dependency from stakwork/strut; remove ./strut

Evanfeenstra:
strut now lives in its own repo, [stakwork/strut](https://github.com/stakwork/strut), with its full history carried over (83 commits back to the first `vein` commit, paths rewritten to the repo root, PR refs rewritten to `stakwork/stakgraph#N`). This PR makes mcp consume it from there and removes `./strut`.

## How mcp depends on it now

```json
"strut": "git+https://github.com/stakwork/strut.git#<commit>"
```

`yarn install` clones the pinned commit, installs strut's devDeps, runs its new `prepare` script (`tsc` + `web/` vite build), and packs by strut's `files` field, so `node_modules/strut` gets `build/` and `web/dist` with no npm release. ~15s locally.

**Do not switch to the GitHub shorthand** (`stakwork/strut#sha`): yarn 1 resolves those to a codeload source tarball and skips `prepare`. Verified both ways in a scratch project.

**To bump strut:** edit the commit in `mcp/package.json`, `yarn install`, commit `yarn.lock`.

**Local dev loop** (replaces `refresh-strut`): `yarn link` in a strut checkout, `yarn link strut` in `mcp/`. yarn keeps the symlink across installs. Documented in `mcp/src/lab/AGENTS.md`.

## Changes
- `mcp/package.json`: git dep; `dev` is plain `tsx src/index.ts`; `refresh-strut` and `stage-strut` removed
- `mcp/Dockerfile`: the `.strut` staging and the `/usr/src/strut` build block are gone; `yarn` builds strut in-image (the image already has `git`)
- CI: `e2e-test` and `e2e-standalone` drop the "Build strut" step; `publish-mcp` drops both rsync staging steps; `strut-desktop.yml` deleted here (it lives in the strut repo now)
- `.gitignore`, `.dockerignore`, `mcp/.gitignore`: strut entries removed
- `mcp/src/lab/AGENTS.md`: dev-loop paragraph rewritten; spec pointers follow the move to `specs/` in the strut repo

## Verified
- `yarn install` in `mcp/` with a clean strut cache: prepare ran, `build/index.js`, `build/index.d.ts`, `web/dist/index.html` present; `yarn.lock` diff is the one strut entry
- `tsc --noEmit` passes; runtime `import("strut")` resolves all 119 exports

## Not verified
- The Docker image build was not run locally. The only behavioural change there is that `yarn` now needs network access to github.com during the build.

## CI fix (second commit): `mcp/.yarnrc` with `network-concurrency 1`

The first push failed both e2e workflows during "Fetching packages" with cache corruption on `@img/sharp-*` (e2e-test) and `rxjs` (e2e-standalone). Cause, confirmed in yarn 1.22's source: a git dependency with a `prepare` script is installed by running a **nested `yarn install`** inside the outer one, both against the same cache, and yarn deletes a package's cache entry before fetching it without holding a lock. On a cold cache with parallel fetches the two installs collide on the packages they share. It never showed locally because my cache was warm.

Reproduced against an empty cache with the branch manifests: fails in 19s with the same errors. With `network-concurrency 1` the identical install passes (73s, `build/` and `web/dist` present). The setting lives in `mcp/.yarnrc` so CI, the Docker build and fresh dev machines all get it; the Dockerfile copies it in before `yarn`. Warm installs are unaffected. Do not remove it.

## Docker verified (third commit): the image builds and `/lab` works

Building the branch's image locally exposed one more thing: the Dockerfile's `npm install --force --os=linux --cpu=x64 @anush008/tokenizers-linux-x64-gnu` step makes npm re-resolve the whole tree, and with strut as a git dep npm re-cloned strut and re-ran its `prepare` its own way. That nested build failed (the `--os/--cpu` flags leak into it, so vite got the wrong rollup binary on arm64), and even when it succeeds it would replace yarn's strut with npm's. The step now uses `npm pack` + `tar`, which downloads that one tarball and leaves `node_modules` exactly as yarn produced it. The onnxruntime-node duplicate check stays and passes.

Then booted the image against a throwaway Neo4j with placeholder API keys:
- `GET /lab/health` → `{"ok":true,"stepCount":127}`
- `GET /lab/` → the strut UI (`<title>strut</title>`), its 876KB bundle served from `node_modules/strut/web/dist`
- `GET /lab/workflows` → 33 seeded workflows; the graph workspace wrote to Neo4j
- no errors in the container log

Inside the image strut's install is identical to a local one: same hoisted versions (`ai` 6.0.175, `@ai-sdk/openai` 3.0.61, `@ai-sdk/anthropic` 3.0.105, one `onnxruntime-node` 1.24.3), same five packages nested under `node_modules/strut`. Those `ai`/`@ai-sdk/openai` versions are older than strut's own ranges because mcp's `resolutions` win; strut's full test suite (709 tests) passes against mcp's tree, so this is a known, tested difference rather than a problem. Previously the image gave strut its own `node_modules` at its own versions.

**Pre-existing, unrelated:** the arm64 image cannot boot at all, before or after this PR. mcp's lockfile only has the darwin, linux-x64 and win32 tokenizers bindings, so on arm64 fastembed's `require('@anush008/tokenizers-linux-arm64-gnu')` throws at startup. For this test I stubbed that module inside the container. Prod is amd64.


Evanfeenstra:
Heads-up: this merged ~18 min before the Dockerfile fix landed on the branch, so `main` currently has a Dockerfile that fails to build (the tokenizers `npm install` step re-resolves the tree and rebuilds strut with npm). The fix is #1672.

PR #1700, opened 2026-09-16
gateway ui: page the run call log instead of truncating at 50

fayekelmith:


Evanfeenstra:
Tested locally (PR UI against a mock run with 137 calls): paging works — 1–50 / 51–100 / 101–137, no gaps or dupes, Prev/Next disable correctly, page 1 shares the header query. 👍

A few things:

1. **"First seen" is wrong for runs > 50 calls.** It reads the oldest row of the first page, so on the 137-call run it showed call #88's time instead of the run start. Pre-existing, but this PR is exactly the case where it shows up.
2. **Table flashes "Loading…" on every page change** (page collapses then re-expands). `placeholderData: keepPreviousData` in `useRunDetail` would fix it.
3. **Nit:** consider paging by `pagination.total_count` (currently dropped by the gateway) instead of `stats.total_requests` — I think Bifrost's stats may exclude in-flight calls, though I didn't verify.

